PR #8, opened 2026-04-09
fix: prevent external import map collisions (#2)

### LINKED ISSUE #2: Transitive external imports can be overwritten by simple-name collisions
## Summary
`_externalImportMap` is keyed by simple name and is overwritten while discovering `[Import]` declarations from referenced assemblies. This makes external import resolution non-deterministic when two libraries expose bindings under the same TS name.

## Evidence
- `MetaSharp.Compiler.TypeScript/Transformation/TypeTransformer.cs:323` does `_externalImportMap[type.Name] = entry`.
- `MetaSharp.Compiler.TypeScript/Transformation/TypeTransformer.cs:327` does `_externalImportMap[tsName] = entry`.
- The comment immediately above says this phase should only add entries and should not overwrite local ones, but the implementation overwrites unconditionally.

## Risk
A transitive dependency can silently replace a local external import mapping. The result depends on `Compilation.References` ordering, which makes the behavior hard to predict and debug.

## Recommendation
Replace simple-name indexing with symbol-qualified or origin-qualified resolution, or at minimum detect collisions and emit a diagnostic. Local entries should explicitly win over transitive ones.

## Missing tests
- Two referenced assemblies with `[Import]` for the same simple name.
- One local assembly and one referenced assembly defining the same external name.
- A collision involving `AsDefault = true` and a named import.

danfma:
## Summary

- Replace unconditional `_externalImportMap[key] = entry` with `RegisterExternalImportMapping` that keeps the first registered entry and emits an `AmbiguousConstruct` warning on collision
- Add `CompileLibrary(source, assemblyName)` and `CompileConsumer(source, params compilations)` overloads to `TranspileHelper` to support multi-library test scenarios
- Add two tests: local-wins-over-referenced collision, and referenced-vs-referenced collision with diagnostic

## Test plan

- [x] All 308 TUnit tests pass
- [x] `LocalImportMapping_WinsOverReferencedCollision` — local `[Import]` wins, diagnostic emitted
- [x] `ReferencedImportCollision_EmitsDiagnosticAndKeepsFirstMapping` — first referenced library wins, diagnostic emitted

Closes #2

### DIFF
--- a/MetaSharp.Compiler.TypeScript/Transformation/TypeTransformer.cs
+++ b/MetaSharp.Compiler.TypeScript/Transformation/TypeTransformer.cs
@@ -116,10 +116,18 @@ public sealed class TypeTransformer(Compilation compilation)
             if (import is not null)
             {
                 var entry = (import.Name, import.From, import.AsDefault, import.Version);
-                _externalImportMap[t.Name] = entry;
+                RegisterExternalImportMapping(
+                    t.Name,
+                    entry,
+                    t.ToDisplayString(),
+                    t.Locations.FirstOrDefault());
                 var tsName = GetTsTypeName(t);
                 if (tsName != t.Name)
-                    _externalImportMap[tsName] = entry;
+                    RegisterExternalImportMapping(
+                        tsName,
+                        entry,
+                        t.ToDisplayString(),
+                        t.Locations.FirstOrDefault());
             }
         }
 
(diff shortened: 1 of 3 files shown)


danfma (commented):
Re: Copilot's comment on `ReferencedImportCollision_EmitsDiagnosticAndKeepsFirstMapping` — the first test (`LocalImportMapping_WinsOverReferencedCollision`) already validates that the winning mapping appears in the generated output and the loser doesn't. This second test focuses specifically on the referenced-vs-referenced scenario where the primary concern is collision detection — the diagnostic is the meaningful assertion here.

Issue #100, opened 2026-04-23
Introduce [Inline] attribute for fields and methods


danfma:
Part of #96. Depends on #96.

## Scope

Introduce `[Inline]` attribute for expansion at use site. Supports fields (initializer inlined) and expression-bodied methods/extensions (body inlined with parameter substitution). Class-level `[Inline]` propagates to all members.

### New attribute: `[Inline]`

- Lives in `Metano.Annotations` (cross-target semantic)
- Multiple targets:
  - **Static readonly field / property**: initializer substitutes at every access site
  - **Expression-bodied method / extension method**: body substitutes at every call site with parameter renaming
  - **Static class**: propagates `[Inline]` to all contained members (ergonomic shortcut)

### Scope limits (explicit)

- Multi-statement method bodies are **out of scope** for this PR (tracked as follow-up) — only expression-bodied methods are inlined
- Recursive inline detection with diagnostic MS00XX `RecursiveInline`
- Closure capture of enclosing state is out of scope (only pure expression forms allowed)

### Pipeline implementation

- IR: `IrField.IsInline`, `IrMethod.IsInline` flags
- Expression bridge: member access to inline field → clone initializer expression tree; call to inline method → clone body with parameter substitution
- Hygienic renaming: temp vars to avoid name collision when inlined body is embedded in caller scope
- Recursion guard: depth limit + cycle detection on inlining pass

### Validation

- MS00XX `InvalidInlineField` — non-static or non-readonly field with `[Inline]`
- MS00XX `InvalidInlineMethod` — non-expression-bodied method with `[Inline]`
- MS00XX `RecursiveInline` — inline expansion cycle detected

## Tests

- Positive: `[Inline]` field with simple literal initializer
- Positive: `[Inline]` field with `new T(...)` initializer + PlainObject lowering
- Positive: `[Inline]` method (extension) with expression body
- Positive: `[Inline]` static class propagates to all fields
- Negative: `[Inline]` on non-readonly field → diagnostic
- Negative: `[Inline]` on multi-statement method → diagnostic
- Negative: recursive inline → MS00XX RecursiveInline
- Integration: DOM bindings sample with full chain (Erasable class + Inline extension + PlainObject record + Inline field) produces expected TS output
- Hygiene: inline body using param name that collides with caller-local gets renamed

## Acceptance

- [ ] `[Inline]` attribute with `AttributeTargets.Field | Property | Method | Class`
- [ ] IR flags and expression bridge updates
[...]

Issue #96, opened 2026-04-23
Attribute family for compile-time erasure and inlining


danfma:
## Summary

Introduce a coherent family of attributes for compile-time erasure, branded wrapping, and inlining. This enables ergonomic DOM-style bindings and literal-type narrowing in the TypeScript target.

## Motivation

The current `[InlineWrapper]` attribute conflates two concerns: branded primitives (runtime-visible distinction) and wrapper inlining. When designing DOM bindings (`document.createElement("div") → HTMLDivElement`), we need:

1. A clean way to erase catalog/container classes at the call site (no runtime footprint)
2. A precise distinction between "provided by runtime" (`[External]`) and ".NET-only" (`[NoEmit]`)
3. Support for inlining field initializers and method bodies at use sites
4. Support for compile-time constant validation on parameters and fields

Today `[ExportedAsModule]` partially covers case (1) but has a latent call-site-flatten bug, and `[NoEmit]` is overloaded to mean both "ambient runtime binding" and ".NET-only type".

## Final attribute family

| Attribute | Target | Purpose |
|-----------|--------|---------|
| `[Erasable]` | static class | Class scope vanishes at call site; members emit per their own attributes |
| `[External]` | class / member | Runtime-provided, no declaration emitted; transpiled code can reference |
| `[NoEmit]` (redefined) | type | .NET-only; diagnostic if referenced from transpiled code |
| `[Branded]` (renames `[InlineWrapper]`) | struct / record | Brand-type wrapper over primitive with companion namespace |
| `[Constant]` | param / field | Value must be compile-time constant; validator enforces literal |
| `[Inline]` | field / method / static class | Expansion at call site (field initializer / method body) |

## Stack plan (4 PRs)

- **PR-0** — `[Erasable]` + refactor `[External]` + redefine `[NoEmit]` + deprecate `[ExportedAsModule]` + migrate DOM bindings — closes the "container" piece
- **PR-1** — `[Constant]` (param + field) with validator MS0013
- **PR-2** — `[Branded]` rename from `[InlineWrapper]`
- **PR-3** — `[Inline]` (field + method/extension + class-propagation)

Each PR ships standalone-useful capability. Stack order chosen so DOM-binding use case works end-to-end only after PR-3 merges.

## Validation scenario

DOM bindings example should transpile as follows after full stack merges:

```csharp
[Erasable]
public static class DocumentExtensions
{
    extension(Document document)
    {
        [Inline]
[...]

PR #105, opened 2026-04-24
feat(core): introduce [Inline] attribute for fields and properties

danfma:
## Summary

Fifth slice of the attribute-family stack (#96 / #100). Introduces `[Inline]` (`Metano.Annotations`) and wires extractor-side substitution: every access to a decorated `static readonly` field or `static` expression-bodied property is replaced by its initializer (or getter expression) before downstream lowering. The declaration itself never materializes in the generated output.

## Changes

**Surface**
- `InlineAttribute` (`AttributeTargets.Field | Property`) with XML docs pinning the two supported shapes and deferring method / extension / static-class propagation to a follow-up slice.
- `DiagnosticCodes.InvalidInline` (MS0016) promoted from reserved to shipped.
- `SymbolHelper.HasInline` with namespace-qualified match against `Metano.Annotations`.

**Extractor** (`IrExpressionExtractor`)
- `ExtractMemberAccess` and `ExtractIdentifierName` now check for `[Inline]` on the resolved symbol and substitute the initializer/getter expression before building an `IrMemberAccess`.
- `_inlineExpanding` (`HashSet<ISymbol>`) guards against cyclic references: reentry on an in-progress member falls back to the regular access path so the substitution stays bounded.
- `TryFindInlineInitializer` resolves the source expression from `DeclaringSyntaxReferences` (`VariableDeclaratorSyntax.Initializer` for fields; `PropertyDeclarationSyntax.ExpressionBody` or the expression-bodied `get` accessor for properties).

**Validator** (`CSharpSourceFrontend.ValidateInlineAttribute`)
- Fields must be `static readonly` with an initializer.
- Properties must be `static` with an expression-bodied getter (property-level `=>` or `get => …`).
- Other targets (methods, events, block-bodied properties) surface MS0016 pointing at the supported shapes.

**DOM-binding use case**
With `[Erasable]` on the container, `[Inline]` on the field, and `[PlainObject]` on the initializer's type, a call like `HtmlElementType.Div` lowers directly to the object literal `{ tagName: "div" }` at every call site — replicating the TypeScript literal-type narrowing pattern without a helper indirection. A dedicated test (`Inline_Property_PlainObjectInitializer_InlinesAsLiteral`) pins the shape.

## Scope boundary

Method / extension-method inlining, static-class `[Inline]` propagation, and an explicit recursion-cycle diagnostic (the extractor currently falls back silently rather than surfacing MS0016 for cycles) ship in a separate follow-up PR. The current slice keeps the lowering path target-agnostic — the attribute's effect lives entirely in the extraction pass.

## Test plan

- [x] `dotnet run --project tests/Metano.Tests/` — **678/678 green** (670 → 678), 8 new tests in `InlineAttributeTranspileTests`:
  - Positive: property literal, field literal, property + `[PlainObject]` initializer, cascading inline through another inline
  - Negative: instance field, mutable field, field without initializer, block-bodied property
- [ ] Reviewer: confirm the cascading-inline case (`Inline_Property_CascadesThroughAnotherInline`) does not regress when the chain grows past 2 levels; recursion-guard depth is O(symbols-in-chain)

## Spec

- `09-attribute-catalog.md`: `InlineAttribute` promoted from *(planned)* to shipped; attribute count 24 → 25. The "planned" footnote is dropped since the family is fully in.
- `10-diagnostic-catalog.md`: MS0016 moved from Reserved Codes to Stable Codes; only MS0013 remains reserved for the `[NoEmit]` redefinition slice.

Part of #96.
Part of #100.

### DIFF
--- a/spec/09-attribute-catalog.md
+++ b/spec/09-attribute-catalog.md
@@ -1,10 +1,9 @@
 # Attribute Catalog
 
 This appendix lists the currently available Metano annotations relevant to the
-transpilation model. The current codebase exposes **24 attribute types** in
+transpilation model. The current codebase exposes **25 attribute types** in
 `Metano.Annotations`, plus the supporting `EmitTarget` enum used by some
-annotations. Additional entries tagged *(planned)* below are reserved for the
-upcoming attribute-family slices (see ADR-0015) and are not yet shipped.
+annotations.
 
 ## Type Selection and Inclusion
 
@@ -45,7 +44,7 @@ upcoming attribute-family slices (see ADR-0015) and are not yet shipped.
 | `DiscriminatorAttribute` (TypeScript) | Names a `[StringEnum]` field as the discriminator; the generated `isT` short-circuits on a literal comparison against the type name before walking the remaining shape. |
 | `ExternalAttribute` (TypeScript) | Marks a `static class` as a stub for runtime-available JS globals — the class emits no file and static member access flattens to a bare identifier. Attribute usage now accepts method/property/field targets so the family can grow; the per-member declaration-suppression lowering and the split from class-level flatten ship in a follow-up slice. |
 | `ConstantAttribute` | Applied to a parameter or field; the value must be a compile-time constant literal. Violations surface as MS0014 `InvalidConstant`. Enables literal-type narrowing in `[Emit]` templates and safe `[Inline]` expansion. |
-| `InlineAttribute` *(planned)* | Will apply to a static readonly field, an expression-bodied method / extension, or a static class to request expansion at every use site. See ADR-0015. |
+| `InlineAttribute` | Applied to a `static readonly` field or a `static` property with an expression-bodied getter; every reference substitutes the member's initializer (or getter expression) at the call site. Method / extension inlining and static-class propagation ship in a follow-up slice. Violations surface as MS0016 `InvalidInline`. |
 
 ## Packaging and Interop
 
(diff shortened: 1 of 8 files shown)


danfma:
@copilot-pull-request-reviewer — thanks for the catch. All 5 threads flagged the same root cause: the `[Inline]` cycle guard was structurally broken because `TryExpandInlineAccess` spawned a nested extractor with a fresh `_inlineExpanding` set every hop, so a transitive `A → B → A` chain would never see the in-progress symbol and could recurse until the stack ran out.

Fixed in `092af05`:
- `IrExpressionExtractor` now shares `_inlineExpanding` across every nested extractor created during inline substitution. The set rides through a private constructor overload so the canonical two-argument public signature stays unchanged. Direct `A => A` and transitive `A => B; B => A` chains bail out at the shared guard.
- `ValidateInlineAttribute.DetectInlineCycles` runs a DFS over every `[Inline]` field/property's initializer (via `DescendantNodesAndSelf()`), following references to other `[Inline]` members until the chain revisits itself. Cycles emit MS0016 with a `"cycle"` keyword in the message *before* extraction runs, so downstream lowering never sees a chain that would recurse.
- MS0016 docstring, `ValidateInlineAttribute` summary, and the `InlineAttribute` XML now describe what the compiler actually does — validator-first detection, shape-specific vs. cycle-specific messages.
- Two regression tests (`Inline_SelfReferentialProperty_EmitsMs0016`, `Inline_MutualRecursion_EmitsMs0016`) pin the cycle diagnostics.

Tests: **680/680 green** (678 → 680). All 5 review threads resolved.

Issue #22, opened 2026-04-11
Root barrel opt-in flag (`--namespace-barrels`)


danfma:
**Background:** [ADR-0006](../blob/main/docs/adr/0006-namespace-first-barrel-imports.md) and [ADR-0007](../blob/main/docs/adr/0007-output-conventions.md) document the current leaf-only barrel strategy. When a package's root namespace is purely sub-namespaces (no types at the bare root), there's no \`src/index.ts\`, and \`import { X } from \"@scope/pkg\"\` without a sub-path doesn't resolve.

The naive fix — emitting \`export * from \"./issues/domain\"; export * from \"./planning/domain\"; …\` at the root — breaks tree-shaking under current bundlers. So the feature is deliberately left as an **opt-in flag** rather than the default.

## Goal

Add a \`--namespace-barrels\` CLI flag that emits a root barrel mirroring the C# namespace hierarchy via \`export namespace\` blocks, rather than \`export *\` aggregation. The generated shape would look like:

\`\`\`ts
// src/index.ts (with --namespace-barrels)
import * as IssuesDomain from \"./issues/domain\";
import * as IssuesApplication from \"./issues/application\";
import * as PlanningDomain from \"./planning/domain\";
import * as SharedKernel from \"./shared-kernel\";

export namespace AcmeIssueTracker {
    export namespace Issues {
        export import Domain = IssuesDomain;
        export import Application = IssuesApplication;
    }
    export namespace Planning {
        export import Domain = PlanningDomain;
    }
    export import SharedKernel = SharedKernel;
}
\`\`\`

Consumers can then write \`import { AcmeIssueTracker } from \"@scope/pkg\"\` and access types as \`AcmeIssueTracker.Issues.Domain.Issue\`. Or keep importing directly from \`#/issues/domain\` as today — both resolve.

## Scope

- [ ] \`--namespace-barrels\` flag on \`metano-typescript\`
- [ ] \`BarrelFileGenerator\` emits the root barrel with \`export namespace\` blocks when the flag is set
- [ ] Preserves tree-shaking: because \`export import\` aliases resolve to specific paths, bundlers can still drop unused branches (verify with a real bundle)
- [ ] \`PackageJsonWriter\` adds the root \`index.ts\` to \`exports[\".\"]\` when the flag is set
- [ ] Tests: transpile a project with nested namespaces, assert the root barrel content + the consumer can import from the package root
[...]

PR #112, opened 2026-04-24
feat(ts): --strip-interface-prefix opt-in for C# interface names

### LINKED ISSUE #92: Drop C# 'I' prefix from generated TS interface names
**Background:** C# style conventions prefix interfaces with `I` (`IIssueRepository`, `IDisposable`). TypeScript community conventions discourage Hungarian notation — most style guides (Google TS, Airbnb, Microsoft's own TS guidelines) and linters (Biome, ESLint via `@typescript-eslint/naming-convention`) flag or forbid it. Metano currently preserves the prefix 1:1, so:

```
// C#
public interface IIssueRepository { … }

// Generated TS
// file: i-issue-repository.ts
export interface IIssueRepository { … }

// Consumer
import { IIssueRepository } from "#/issues/application/i-issue-repository";
```

A consumer who mixes Metano output with hand-written TS (React, Hono, effect, etc.) ends up with an inconsistent naming surface. The `I` prefix is the one visual noise that reliably marks "this came from C#".

## Goal

Add a mechanism to drop the `I` prefix in generated TS output when:

- The C# interface name matches `/^I[A-Z]/` (captures `IIssueRepository` but leaves `Identity`, `ISO8601` alone).
- Stripping the prefix would **not** collide with another top-level type in the same namespace (e.g., if the project ships both an `IIssueRepository` interface and an unrelated `IssueRepository` class, both would want to emit as `IssueRepository`).

## Proposed design — opt-in flag (Option A)

Ship as an opt-in `--strip-interface-prefix` CLI flag plus matching `MetanoStripInterfacePrefix` MSBuild property, following the same pattern as `--namespace-barrels` (#22). Default stays `false` so existing consumers see no change on upgrade.

Emission matrix when the flag is set:

| C# | TS | Notes |
|---|---|---|
| `IIssueRepository` (no sibling) | `IssueRepository` | prefix dropped |
| `IIssueRepository` + `IssueRepository` class in same namespace | `IIssueRepository` + `IssueRepository` | collision → keep prefix, emit `MS0012` warning |
| `IIssueRepository` + `[Name(TypeScript, "Custom")]` | `Custom` | explicit override wins, prefix logic skipped |
| `Identity` (no uppercase second char) | `Identity` | unchanged |
| `ISO8601Spec` (`I` + `S` uppercase — would capture) | ambiguous — see below | |

**Open question**: should the heuristic be `/^I[A-Z]/` (captures both `IFoo` and `ISO8601`) or stricter (`/^I[A-Z][a-z]/` — requires Pascal-case sequel)? The stricter form preserves acronym-prefixed names but gets spoofed by `IOrder` (genuine interface). **Recommendation**: `/^I[A-Z]/` — matches C# IDE refactor heuristics. Users who want to preserve a specific non-interface name use `[Name(TypeScript, "…")]`.

## Alternative — default flip (Option B)

Flip the behavior by default and emit MS0012 for collisions, with `[Name(TypeScript, "IFoo")]` as the explicit escape hatch. Cleaner outcome but breaking change — every existing consumer's imports need updating. A major version bump is the right vehicle; holds until the project cuts one.

## Scope (if Option A is accepted)

- [ ] `MetanoStripInterfacePrefix` attribute / CLI flag / MSBuild property (routed through `Metano.Build.targets`)
- [ ] Frontend pass: compute the stripped name, detect collisions in the same namespace (reusing `TranspilableTypes` projection from #56 Phase B.10)
- [ ] Update file-name derivation (`i-issue-repository.ts` → `issue-repository.ts`) when the strip applies
- [ ] Update cross-assembly import resolution so consumer packages also see the stripped name (`ir.TypeNamesBySymbol` entries reflect the strip)
- [ ] `MS0012` diagnostic for collisions; include both colliding type names and the namespace in the message
- [ ] Tests covering: strip emission, collision keep-prefix path, `[Name]` override takes precedence, stripped name in cross-assembly imports
- [ ] Sample: flip `SampleIssueTracker.csproj` into the opt-in once the mechanics land so generated output demonstrates the shape; update `test/issues/application/issue-service.test.ts` import paths and Copilot can verify Biome stays quiet
- [ ] ADR update covering the convention + the rationale for opt-in default

## Out of scope

- Making it the default. Bundled with a major version bump once the project decides to cut one.
- Any interaction with C# abstract classes (no prefix convention exists there).
- Stripping other Hungarian notations (`T` for type-parameters is TS-idiomatic, not a transpiler choice).

## Rationale for opt-in default

Same argument as #22 (`--namespace-barrels`): consumer import surface is a compatibility boundary. Flipping defaults invalidates every downstream import line in one commit. An opt-in lets individual projects migrate on their own schedule, and accumulates real-world friction data before the default flip.

Part of #14.

danfma:
## Summary

Closes #92 (mechanics). Opt-in `--strip-interface-prefix` CLI flag (also surfaced as `MetanoStripInterfacePrefix` MSBuild property) rewrites every transpilable interface whose C# name matches `^I[A-Z]` to drop the leading `I` when emitted to TypeScript:

```
public interface IIssueRepository { ... }
// default:  file i-issue-repository.ts, `export interface IIssueRepository`
// opt-in:   file issue-repository.ts,   `export interface IssueRepository`
```

Default stays off — same compatibility story as `--namespace-barrels` (#22). Flipping by default waits for a major version bump.

## Pipeline

- `Commands.Transpile` accepts the new `stripInterfacePrefix` flag and forwards it to `TypeScriptTarget`.
- `TypeScriptTarget.StripInterfacePrefix` forwards into `TypeTransformer`.
- `TypeTransformer.TransformAll` clones `ir.TypeNamesBySymbol` into a local mutable dictionary and runs `ApplyInterfacePrefixStrip` once. The pass indexes transpilable types by namespace, detects collisions with sibling types, and rewrites the entry in place.
- `ResolveTsName`, file-name derivation (`GroupTypesByFile`), and `ImportCollector` all consume the rewritten map transparently — no per-consumer plumbing.
- `IrToTsInterfaceBridge.Convert` accepts an optional `nameOverride` so the emitted `export interface <Name>` matches the rewritten dictionary entry.
- `Metano.Build.targets` wires the MSBuild property into `_MetanoArgs`.

## Collision handling (MS0017)

When the stripped name would clash with another top-level type in the same namespace, the transformer keeps the prefixed name and emits `MS0017 InterfacePrefixCollision` as a warning. Resolution: rename the conflicting type, or set `[Name(TypeScript, "…")]` on one side.

## `[Name]` override precedence

Author-provided `[Name(TypeScript, "…")]` wins over the strip. The pass skips entries whose current (frontend-registered) name diverges from the Roslyn type name, so the author-chosen TS name passes through unchanged.

## Out of scope

- Sample migration (`SampleIssueTracker.csproj` flipping to the opt-in) — follow-up.
- ADR update covering the opt-in default rationale — follow-up.

## Test plan

- [x] `dotnet run --project tests/Metano.Tests/` — **696/696 green** (689 → 696), 6 new tests in `StripInterfacePrefixTests`:
  - `NoCollision_DropsPrefix` — canonical rewrite
  - `OffByDefault_PreservesPrefix` — compat guard
  - `CollidesWithSiblingClass_KeepsPrefixAndEmitsMs0017` — diagnostic + fallback
  - `NameOverride_WinsOverStrip` — author override precedence
  - `IdentityShape_DoesNotMatch` — `/^I[A-Z]/` heuristic guard
  - `DoesNotTouchClasses` — interface-only scope
- [ ] Reviewer: confirm the mutable-copy of `TypeNamesBySymbol` in `TypeTransformer.TransformAll` does not regress any cross-assembly lookup path — the transformer-local copy is what every downstream site reads
- [ ] Reviewer: sanity-check the heuristic scope (`/^I[A-Z]/` captures `IFoo` and `ISO8601` alike; users opt out per-type via `[Name(TypeScript, "…")]`)

Closes #92.

### DIFF
--- a/spec/10-diagnostic-catalog.md
+++ b/spec/10-diagnostic-catalog.md
@@ -12,7 +12,7 @@ Each diagnostic carries:
 - message;
 - optional source location.
 
-The current stable code range is **`MS0001` through `MS0016`**, with
+The current stable code range is **`MS0001` through `MS0017`**, with
 `MS0013` reserved for the upcoming `[NoEmit]` redefinition slice
 (see ADR-0015).
 
@@ -35,6 +35,7 @@ The current stable code range is **`MS0001` through `MS0016`**, with
 | `MS0014` | `InvalidConstant` | `[Constant]` argument or initializer is not a compile-time constant literal. |
 | `MS0015` | `InvalidErasable` | `[Erasable]` was applied to a non-static class, or combined with `[Transpile]`. |
 | `MS0016` | `InvalidInline` | `[Inline]` was applied to an unsupported shape (instance or mutable field, field without initializer, block-bodied property, non-static property, or any other target). |
+| `MS0017` | `InterfacePrefixCollision` | Stripping the `I` prefix from an interface name (under `--strip-interface-prefix`) would collide with another top-level type in the same namespace. Keeps the prefix so the consumer surface stays compilable. |
 
 ## Reserved Codes
 
--- a/src/Metano.Build/build/Metano.Build.targets
+++ b/src/Metano.Build/build/Metano.Build.targets
@@ -48,6 +48,9 @@
       <_MetanoArgs Condition="'$(MetanoNamespaceBarrels)' == 'true'"
         >$(_MetanoArgs) --namespace-barrels</_MetanoArgs
       >
+      <_MetanoArgs Condition="'$(MetanoStripInterfacePrefix)' == 'true'"
+        >$(_MetanoArgs) --strip-interface-prefix</_MetanoArgs
+      >
       <_MetanoArgs Condition="'$(MetanoExtraArgs)' != ''"
         >$(_MetanoArgs) $(MetanoExtraArgs)</_MetanoArgs
       >
--- a/src/Metano.Build/build/Metano.Build.targets
+++ b/src/Metano.Build/build/Metano.Build.targets
@@ -16,6 +16,8 @@
       MetanoSrcRoot          (optional)  TS source root relative to package root (default: inferred)
       MetanoSkipPackageJson  (optional)  true to skip package.json generation (default: false)
       MetanoShowTimings      (optional)  true to print timing info (default: false)
+      MetanoNamespaceBarrels (optional)  true to emit a root src/index.ts barrel mirroring C# namespaces (default: false)
+      MetanoStripInterfacePrefix (optional)  true to drop the `I` prefix from interface names when the second letter is uppercase (default: false). Collisions with sibling types surface MS0017.
       MetanoExtraArgs        (optional)  Additional CLI arguments
   -->
 
--- a/src/Metano.Compiler.TypeScript/Bridge/IrToTsInterfaceBridge.cs
+++ b/src/Metano.Compiler.TypeScript/Bridge/IrToTsInterfaceBridge.cs
@@ -9,7 +9,11 @@ namespace Metano.TypeScript.Bridge;
 /// </summary>
 public static class IrToTsInterfaceBridge
 {
-    public static void Convert(IrInterfaceDeclaration ir, List<TsTopLevel> statements)
+    public static void Convert(
+        IrInterfaceDeclaration ir,
+        List<TsTopLevel> statements,
+        string? nameOverride = null
+    )
     {
         var properties = new List<TsProperty>();
         var methods = new List<TsInterfaceMethod>();
@@ -34,7 +38,12 @@ public static class IrToTsInterfaceBridge
             }
         }
 
-        var tsName = IrToTsNamingPolicy.ToTypeName(ir.Name, ir.Attributes);
+        // A transformer-provided override (e.g.,
+        // `--strip-interface-prefix`) takes precedence over the naming
+        // policy so the emitted declaration matches the name the
+        // transformer already registered in `TypeNamesBySymbol` (and
+        // consequently the file name + import collector).
+        var tsName = nameOverride ?? IrToTsNamingPolicy.ToTypeName(ir.Name, ir.Attributes);
         var typeParams = ConvertTypeParameters(ir.TypeParameters);
 
         statements.Add(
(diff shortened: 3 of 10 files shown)


danfma:
@gemini-code-assist @copilot-pull-request-reviewer — all six threads addressed in `1b4d5c4`. Summary per gap:

- **Cross-file references not rewritten (Copilot P1)**: `IrToTsTypeMapper.MapNamed` now consults an `AsyncLocal<IReadOnlyDictionary<string, string>?>` rename slot that `TypeTransformer.TransformAll` publishes for the duration of each run. Callers who never passed custom `IrToTsTypeOverrides` still pick up the rewrite because the check lives on the named-type path itself. `AsyncLocal` prevents the race that a plain static field would introduce under TUnit's parallel test execution. New `StripInterfacePrefix_ConsumerImportsUseStrippedName` pins the end-to-end case: a `Service` class referencing `IIssueRepository` now renders `IssueRepository` as parameter + property type and imports from `./issue-repository`.
- **`TranspilableTypes` / `TsName` / `FileName` left stale (Copilot)**: new `RewriteTranspilableTypeEntry` mutates the cloned `TranspilableTypes` dict (TsName + FileName + key swap) alongside the `typeNamesBySymbol` rewrite, so the import collector + guard resolver see the stripped form consistently.
- **Nested-interface false collisions (Gemini HIGH)**: `GetEnclosingScopeKey` returns the containing type for nested interfaces and the namespace display for top-level types. The sibling index no longer mixes interfaces from unrelated outer types that happen to share a namespace.
- **Scope limit for nested interfaces (Copilot XML docs)**: the strip explicitly skips `type.ContainingType is not null` in this slice and the docs call that out — the bridge + file-naming paths for nested types have not been audited for the rewritten identifier yet, so a mixed shape (stripped outer + prefixed nested) would surprise consumers. Tracked for follow-up.
- **Global namespace rendering (Copilot)**: MS0017 messages now read "in the global namespace" or "in namespace 'X'" instead of the previous `in ''`.
- **`MetanoStripInterfacePrefix` missing from targets header (Copilot)**: the header comment in `Metano.Build.targets` now lists both `MetanoNamespaceBarrels` and `MetanoStripInterfacePrefix` so consumers discover the MSBuild switches without spelunking the file.

Tests: **697/697 green** (696 → 697). All 6 review threads resolved.

Issue #113, opened 2026-04-24
Introduce [This] attribute to model JS function `this` context


danfma:
## Motivation

Several JS/TS APIs rely on the function's `this` binding rather than an explicit parameter. The canonical case is a DOM event handler:

```ts
element.onclick = function (event) {
  this.innerHTML = "clicked";
};
```

Here `this` is the element that received the event, and the function must be declared with the `function` keyword so the runtime can rebind `this` at dispatch. TypeScript models the contract through a synthetic first parameter:

```ts
type MouseEventListener = (this: Element, event: MouseEvent) => void;
```

C# has no native way to express "the first parameter is the receiver" on a delegate or an ordinary method. Our DOM binding currently has to either drop `this` (losing the type information) or expose it as a plain first parameter, which then emits incorrectly as a regular argument.

Concrete need from `bindings/Metano.TypeScript.DOM/Element.cs`:

```csharp
[Name("onclick")]
public MouseEventListener? OnClick { get; set; }

public delegate void MouseEventListener(MouseEvent @event);
```

The author wants to write handlers that access the element via `this`, but C# provides no first-class way to declare that intent.

## Proposal

Introduce `[This]` (`Metano.Annotations.ThisAttribute`) — a parameter-level attribute that marks the first parameter of a delegate or inlinable method as the synthetic `this` receiver.

```csharp
public delegate void MouseEventListener([This] Element self, MouseEvent @event);
```

Lowering contract:

- The decorated parameter is removed from the emitted TS parameter list.
- TypeScript's synthetic `this` parameter is inserted with the same type: `(this: Element, event: MouseEvent) => void`.
- Identifier references to the decorated parameter inside the body are rewritten to `this` (property access chains included).
- Any function literal assigned to a delegate with a `[This]` first parameter must emit with the `function` keyword instead of an arrow expression, so the runtime `this` rebinds correctly.

### Delegate declaration

```csharp
public delegate void MouseEventListener([This] Element self, MouseEvent @event);
```

```ts
export type MouseEventListener = (this: Element, event: MouseEvent) => void;
```

### Lambda / anonymous-method assignment

```csharp
button.OnClick = (self, e) => self.InnerHtml = "clicked";
```

Because C# lambda parameters cannot carry attributes, the `[This]` painting is inherited from the target delegate's first parameter. Emission:

```ts
button.onclick = function (e) {
  this.innerHTML = "clicked";
};
```

[...]

PR #114, opened 2026-04-24
feat(ts): introduce [This] attribute — Slice A (delegate type signature)

danfma:
## Summary

First slice of #113. Introduces `ThisAttribute` in `Metano.Annotations` and teaches the frontend to hoist a decorated first parameter into the IR function type's synthetic `this` receiver. The TypeScript printer emits `(this: T, …) => R` when the slot is populated; Dart (and any future non-JS backend) keeps the parameter positional by falling through to its own lowering path (no-op today).

PoC before committing scope verified that Roslyn's `SemanticModel.GetTypeInfo(lambda).ConvertedType` exposes the target delegate symbol and that the delegate's first-parameter attributes round-trip through the compilation — the building block Slice B will lean on for lambda target-type inference.

## What this slice covers

- `ThisAttribute` (`AttributeTargets.Parameter`) with XML docs pinning the first-parameter + delegate-or-inlinable-method contract.
- `DiagnosticCodes.InvalidThis` (MS0018) promoted to Stable Codes.
- `SymbolHelper.HasThis` with namespace-qualified match against `Metano.Annotations`.
- Frontend validator `ValidateThisAttribute` walks every method + delegate invoke method and surfaces MS0018 when:
  - the decorated parameter is not at index 0, or
  - the parameter carries `ref` / `out` / `params`.
- `IrFunctionTypeRef` gains an optional `ThisType` slot. The extractor (`IrTypeRefMapper.MapDelegateType`) promotes the first parameter into the slot and removes it from the positional list when `HasThis` fires.
- `TsFunctionType` gains a matching `ThisType`; the TS printer emits `(this: T, param: U) => R` (or `(this: T) => R` when no trailing params remain). The bridge wires the IR → TS copy.

## What lands later

- **Slice B** — body rewriting: lambda target-type inference publishes `uses-this`; the printer picks `function (…)` over `=>`; identifier references to the `[This]` parameter rewrite to `this`.
- **Slice C** — method-group assignment, explicit-receiver invocation (`.call(receiver, …)`), runtime interop for delegate `+=` / `-=`.
- **Slice D** — cross-assembly delegate resolution + Dart no-op bridge + end-to-end DOM binding sample + ADR.

## Test plan

- [x] `dotnet run --project tests/Metano.Tests/` — **702/702 green** (697 → 702), 5 new tests in `ThisAttributeTranspileTests`:
  - `This_OnDelegateFirstParameter_EmitsThisAnnotationInFunctionType` pins `(this: Element, arg: string) => void`
  - `This_OnNonFirstParameter_EmitsMs0018`
  - `This_OnRefParameter_EmitsMs0018`
  - `This_OnParamsParameter_EmitsMs0018`
  - `This_OnSingleParameterDelegate_EmitsEmptyParameterList` pins `(this: Element) => void` (no trailing comma)
- [ ] Reviewer: confirm MS0018 messaging is precise enough for adoption (non-first-position vs. `ref`/`out`/`params`)
- [ ] Reviewer: sanity-check the `IrFunctionTypeRef.ThisType` shape before Slice B threads it into body rewriting

## Spec

- `09-attribute-catalog.md`: attribute count 25 → 26; `ThisAttribute` row added with the Slice A contract (signature emission only; body rewriting deferred).
- `10-diagnostic-catalog.md`: MS0018 promoted from Reserved to Stable (only MS0013 remains reserved, for the `[NoEmit]` redefinition slice).

Part of #113.

### DIFF
--- a/spec/09-attribute-catalog.md
+++ b/spec/09-attribute-catalog.md
@@ -1,7 +1,7 @@
 # Attribute Catalog
 
 This appendix lists the currently available Metano annotations relevant to the
-transpilation model. The current codebase exposes **25 attribute types** in
+transpilation model. The current codebase exposes **26 attribute types** in
 `Metano.Annotations`, plus the supporting `EmitTarget` enum used by some
 annotations.
 
@@ -45,6 +45,7 @@ annotations.
 | `ExternalAttribute` (TypeScript) | Marks a `static class` as a stub for runtime-available JS globals — the class emits no file and static member access flattens to a bare identifier. Attribute usage now accepts method/property/field targets so the family can grow; the per-member declaration-suppression lowering and the split from class-level flatten ship in a follow-up slice. |
 | `ConstantAttribute` | Applied to a parameter or field; the value must be a compile-time constant literal. Violations surface as MS0014 `InvalidConstant`. Enables literal-type narrowing in `[Emit]` templates and safe `[Inline]` expansion. |
 | `InlineAttribute` | Applied to a `static readonly` field or a `static` property with an expression-bodied getter; every reference substitutes the member's initializer (or getter expression) at the call site. Method / extension inlining and static-class propagation ship in a follow-up slice. Violations surface as MS0016 `InvalidInline`. |
+| `ThisAttribute` | Applied to the first parameter of a delegate or inlinable method; promotes the slot to the synthetic JavaScript `this` receiver. The parameter is dropped from the emitted TS positional list and re-introduced as the function type's `this` annotation (`(this: T, …) => R`). Lambda / method-group / body-rewrite emission lands in a follow-up slice. Violations surface as MS0018 `InvalidThis`. |
 
 ## Packaging and Interop
 
--- a/spec/10-diagnostic-catalog.md
+++ b/spec/10-diagnostic-catalog.md
@@ -12,7 +12,7 @@ Each diagnostic carries:
 - message;
 - optional source location.
 
-The current stable code range is **`MS0001` through `MS0017`**, with
+The current stable code range is **`MS0001` through `MS0018`**, with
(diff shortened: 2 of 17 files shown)


danfma:
@gemini-code-assist @copilot-pull-request-reviewer — all four threads addressed in `1e80d6e`.

- **`in` parameter rejection (Gemini + Copilot duplicate)**: the validator now rejects `RefKind.In` alongside `Ref` / `Out` / `RefReadOnly` / `params`. The MS0018 message lists `in` explicitly and a new test `This_OnInParameter_EmitsMs0018` pins the behavior.
- **Dart bridge silently dropped the receiver (Copilot P1)**: `IrToDartTypeMapper` re-introduces `ThisType` as a regular positional parameter at index 0 when the slot is populated. Dart has no JS-style receiver rebinding, so the attribute degrades to a no-op rather than losing the type from the signature. New `DartBackendTests.ThisAttribute_DartBridge_ReintroducesReceiverAsFirstParameter` pins that the Dart output still carries the receiver type.
- **Walkers skipped `ThisType` (Copilot)**: `IrRuntimeRequirementScanner` and the TS class emitter's `CollectTypeParamRefs` now recurse into `f.ThisType` alongside `f.Parameters` and `f.ReturnType`. Future generic receivers (type parameter, Temporal subtype, `[EmitPackage]`-sourced type) correctly contribute to import collection and type-parameter inference.
- **`IrTypeRefSignatures` did not distinguish `this:`-bearing functions (Copilot)**: the descriptor now prefixes `this:T,` inside the `fn<…->R>` key when ThisType is set, so downstream signature-based deduplication does not collapse a `(this: Element, string) => void` onto the shape-compatible `(Element, string) => void`.

Tests: **704/704 green** (702 → 704). All 4 review threads resolved.

PR #123, opened 2026-04-24
feat(ts): [This] Slice B — lambda binding and body rewrite

danfma:
## Summary

Second slice of #113. Teaches the extractor to detect lambdas bound to a `[This]`-bearing delegate, drop the lambda's first parameter from the emitted positional list, rewrite identifier references to the dropped parameter as `IrThisExpression`, and tell the TypeScript printer to emit the `function` keyword so the runtime rebinds `this` at dispatch. Arrow functions capture the lexical `this` and would defeat the receiver contract, so the two forms are explicitly split at the printer level.

## Extractor

- New `_thisParameterSymbols` set carries lambda parameters promoted to JS `this`. Shared across nested extractors so a block-bodied lambda body rewrites through the same mechanism as an expression-bodied one — `IrStatementExtractor` gains an internal constructor that shares its expression extractor with the caller.
- `ExtractIdentifierName` short-circuits to `IrThisExpression` when the resolved symbol sits in `_thisParameterSymbols`. Member access falls through the existing `Extract(target)` path, so `self.InnerHtml` naturally becomes `this.innerHtml`.
- `ResolveLambdaThisBinding` uses `SemanticModel.GetTypeInfo(lambda).ConvertedType` to find the target delegate. When the delegate's first parameter carries `[This]`, the extractor records the matching lambda-parameter symbol in the set for the duration of body extraction, drops the parameter from the IR's positional list, and stamps `IrLambdaExpression.UsesThis` / `ThisType`.

## IR

- `IrLambdaExpression` gains `UsesThis` + `ThisType` slots (both default-false / null). Other backends ignore them.

## TypeScript bridge + printer

- `IrToTsExpressionBridge.MapLambda` forwards `ThisType` to the TS AST.
- `TsArrowFunction.ThisType` holds the receiver type; the printer switches to the classic `function (this: T, …) { … }` shape when the slot is populated. Existing arrow emission stays the default for plain delegates.

## Scope boundary

Deferred to later slices:

- **Slice C** — method-group assignment (direct reference + `.call(receiver, …)` wrapper on shape mismatch); explicit-receiver delegate invocation from C#; named-method declarations carrying `[This]` on their own first parameter.
- **Slice D** — Dart backend no-op confirmation (already in place via #114 review fixes) + cross-assembly delegate pickup + end-to-end DOM binding sample + ADR.

## Test plan

- [x] `dotnet run --project tests/Metano.Tests/` — **707/707 green** (704 → 707), 3 new tests in `ThisAttributeTranspileTests`:
  - `This_LambdaAssignedToDelegate_EmitsFunctionKeywordAndThisReceiver` pins the canonical `function (this: Element, arg: string) { … }` shape and the body rewrite (`self.InnerHtml` → `this.innerHtml`).
  - `This_LambdaWithoutTargetDelegate_StillEmitsArrow` regression-guards that ordinary `Action<T>` targets keep the arrow lowering.
  - `This_SingleParameterLambda_PromotesEntireParameterToThis` pins the `self => …` shape where the only parameter is the receiver (empty positional list).
- [ ] Reviewer: confirm the identifier-rewrite path does not over-reach (only `IParameterSymbol` with the registered symbol rewrites; method parameters / local variables with the same name do not)
- [ ] Reviewer: sanity-check the `TsArrowFunction.ThisType` branch in the printer; async + this combination should still render `async function (this: T, …) { … }`

Part of #113.

### DIFF
--- a/src/Metano.Compiler.TypeScript/Bridge/IrToTsExpressionBridge.cs
+++ b/src/Metano.Compiler.TypeScript/Bridge/IrToTsExpressionBridge.cs
@@ -370,7 +370,20 @@ public static class IrToTsExpressionBridge
             ))
             .ToList();
         var body = IrToTsStatementBridge.MapBody(lambda.Body, bclRegistry);
-        return new TsArrowFunction(parameters, body, Async: lambda.IsAsync);
+        var arrow = new TsArrowFunction(parameters, body, Async: lambda.IsAsync);
+        // Lambdas bound to a `[This]`-bearing delegate: wrap the arrow
+        // in a runtime `bindReceiver(...)` call. The helper's
+        // `function`-keyword trampoline picks up the runtime `this`
+        // from the caller and forwards it as the first positional
+        // argument to the arrow — the arrow itself stays
+        // lexically-scoped so any `this` captured from the enclosing
+        // C# class still resolves through ordinary closure
+        // semantics. Zero body rewriting needed; the arrow's
+        // receiver parameter (e.g. `self`) carries the runtime
+        // `this`.
+        if (lambda.UsesThis)
+            return new TsCallExpression(new TsIdentifier("bindReceiver"), [arrow]);
+        return arrow;
     }
 
     /// <summary>
--- a/src/Metano.Compiler.TypeScript/Transformation/ImportCollector.cs
+++ b/src/Metano.Compiler.TypeScript/Transformation/ImportCollector.cs
@@ -852,6 +852,14 @@ public sealed class ImportCollector(
                 {
                     names.Add(callId.Name);
                     valueNames.Add(callId.Name);
+                    // `bindReceiver` is the metano-runtime helper the
+                    // TS bridge emits around lambdas bound to a
+                    // `[This]`-bearing delegate. Route it through the
+                    // runtime-helper bucket so the collector emits
+                    // the matching import without threading a new
+                    // IrRuntimeRequirement through the frontend.
+                    if (callId.Name == "bindReceiver")
+                        runtimeHelpers.Add("bindReceiver");
                 }
                 // Static method calls like Enumerable.from(...)
                 else if (
--- a/src/Metano.Compiler/Extraction/IrExpressionExtractor.cs
+++ b/src/Metano.Compiler/Extraction/IrExpressionExtractor.cs
@@ -40,7 +40,7 @@ public sealed class IrExpressionExtractor
     )
         : this(semanticModel, originResolver, target, inlineExpanding: null) { }
 
-    private IrExpressionExtractor(
+    internal IrExpressionExtractor(
         SemanticModel semanticModel,
         IrTypeOriginResolver? originResolver,
         Metano.Annotations.TargetLanguage? target,
@@ -52,6 +52,8 @@ public sealed class IrExpressionExtractor
         _target = target;
         _inlineExpanding = inlineExpanding ?? new HashSet<ISymbol>(SymbolEqualityComparer.Default);
     }
+
+    internal HashSet<ISymbol> InlineExpandingSet => _inlineExpanding;
 
     public IrExpression Extract(ExpressionSyntax expression) =>
         expression switch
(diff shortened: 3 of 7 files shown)


danfma:
@gemini-code-assist @copilot-pull-request-reviewer — both threads addressed in `1c5a7d0`.

- **Nested `[This]` lambdas aliasing outer receivers (Gemini HIGH + Copilot)**: swapped the `HashSet<ISymbol>` for a single-field save/restore stack on the extractor. Only the innermost `[This]` promotion is eligible for the identifier → `this` rewrite. Outer receivers referenced from inside a nested `[This]` lambda stay under their original C# name (ordinary closure capture). `Register`/`Unregister` helpers removed in favor of try/finally save of `_currentThisParameterSymbol`. New `This_NestedLambdas_InnerReceiverDoesNotAliasOuterReceiver` regression-guards the scope.
- **`async` + `[This]` printer coverage (Copilot)**: new `This_AsyncLambda_EmitsAsyncFunctionKeyword` pins the `async function (this: T, ...) { ... }` keyword ordering for async lambdas bound to a `[This]` delegate.

Tests: **709/709 green** (707 → 709). All 4 review threads resolved.

danfma:
@gemini-code-assist @copilot-pull-request-reviewer — **abordagem substituída**. Refiz a Slice B inteira com uma estratégia diferente após discussão. A revisão anterior identificou corretamente o risco de aliasing em lambdas aninhadas; durante a confirmação descobri um problema adicional (o `function`-keyword shadow alia também o `this` da classe enclosing capturado por closure), e a conclusão foi que o body rewrite era estrutural-mente frágil para o caso geral.

A branch foi force-pushed (o histórico anterior com 2 commits foi substituído por um único commit `feat(ts): [This] Slice B — bindReceiver runtime wrap`). Por favor revisem de novo.

**Resumo da mudança**:

- **Descartado**: emissão de `function (this: T, ...) { ... }` + rewrite de referências ao parâmetro `[This]` como keyword `this`. O shadow da keyword no body entrava em conflito com `this` do outer C# class capturado via closure (ex: `OnClick = self => self.InnerHtml = this.Name` — os dois `this` acabavam apontando pro Element em runtime).
- **Nova abordagem**: a lambda emite como arrow plain wrapped em `bindReceiver(...)`, um helper novo em `metano-runtime/src/system/bind-receiver.ts`. O helper é um `function`-keyword trampoline que captura o `this` do dispatcher e forward como primeiro argumento posicional pra arrow. A arrow mantém `this` lexicamente; o receiver viaja no parâmetro (`self`, etc.). Closure do outer class continua natural.
- **Zero body rewrite** — o extractor não toca mais no body. `_currentThisParameterSymbol`, `IdentifierName → IrThisExpression`, o `IrStatementExtractor` interno constructor, o `TsArrowFunction.ThisType` slot e o printer branch de `function`-keyword foram todos removidos.
- **Auto-import**: `ImportCollector` detecta identificadores `bindReceiver` em call-sites e roteia pro bucket existente de runtime helpers — sem nova categoria de `IrRuntimeRequirement`.

**Fix direto pros dois pontos que vocês levantaram**:
- Nested `[This]` lambdas (Gemini HIGH + Copilot): cada nível wrappa na sua própria chamada `bindReceiver`; as arrows ficam lexicamente isoladas, o outer receiver é uma closure natural sob o nome C# original.
- `async` + `[This]` (Copilot): emite `bindReceiver(async (self, ...) => ...)` — o async qualifier permanece no argumento do helper.

**Testes**: 6 novos em `ThisAttributeTranspileTests` cobrindo canonical + plain-arrow regression + single-param receiver + nested + async + **closure do outer `this`** (motivador principal do redesign). Full TUnit: 704 → 710, 0 failures.

**Issue #113** atualizada com detalhes do design change + acceptance checklist revisto. Se fizer sentido, podem rebaselinar a revisão contra o novo commit. Obrigado pela atenção aos detalhes que apontou o buraco original.

danfma:
@gemini-code-assist — your latest review (21:21 UTC) landed on the pre-force-push diff and is still reviewing the previous `function`-keyword + body-rewrite shape. That approach was replaced at ~`53059f2` (current HEAD). The feedback you gave would be correct for the earlier draft but no longer applies:

- **"Store the original parameter name so the backend can emit `const self = this;`"** — the redesigned slice does not drop the receiver parameter, does not use the `function` keyword on the generated lambda, and does not rewrite any identifier. The arrow keeps the user's original receiver parameter (`self`, or whatever they named it); the runtime helper `bindReceiver` is the classic `function` that picks up the dispatcher's `this` and forwards it as the first argument. Nested `function` callbacks rebinding `this` is a non-issue because the generated lambdas are arrows.
- **"Emit `const self = this;` inside the function body"** — no function body is synthesized. The user's body flows through untouched (including any outer-class `this` reference, which stays a lexical closure capture).

Please re-run the review against `53059f2`. TL;DR of the redesign is in the [issue comment](https://github.com/danfma/metano/issues/113#issuecomment-4316462072) and the [PR comment explaining the force-push](https://github.com/danfma/metano/pull/123#issuecomment-4316471293). The `This_LambdaCapturingOuterClassThis_KeepsLexicalBinding` test pins the exact scenario your review flagged as a potential defect — under the new approach the lambda body reads `self.innerHtml = this.name` correctly, where `self` is the runtime receiver (forwarded by `bindReceiver`) and `this` is the enclosing C# class captured via the arrow's lexical scope.

@copilot-pull-request-reviewer — likewise, your earlier feedback about nested-lambda receiver aliasing no longer applies under `bindReceiver`; the scenario is covered by `This_NestedLambdas_EachReceiverFlowsThroughItsOwnBindReceiver`. A fresh pass against the current diff would be helpful.

danfma:
@gemini-code-assist — both docstring/comment drift issues addressed (amended into the existing squashed commit, ready for force-push):

- **`IrLambdaExpression` XML docs**: rewrote to describe the current `bindReceiver` contract — the receiver parameter stays in `Parameters`, `UsesThis`/`ThisType` signal the bridge to hand the arrow to the helper, arrow keeps lexical `this` for outer-class closure. The old "drops receiver + emits `function` keyword" sentence is gone.
- **`TryExpandInlineAccess` comment**: trimmed the stale "this-parameter tracking stays at defaults" sentence; only the cycle-tracking mention remains, matching what the extractor actually does now.

Tests: **710/710 green** after the edit. Both review threads resolved.

PR #124, opened 2026-04-25
feat(ts): [This] Slice C+D — method group, .call invocation, ADR

### LINKED ISSUE #113: Introduce [This] attribute to model JS function `this` context

danfma:
## Summary

Closes #113. Final two slices of the `[This]` attribute work shipped together:

## Slice C — method-group + direct invocation

- **Method-group assignment** to a `[This]`-bearing delegate wraps the named handler reference in `bindReceiver(...)`. Both instance (`OnClick = Handler`) and static (`OnLoad = Handlers.OnLoad`) groups flow through `WrapMethodGroupForThisDelegate` in `IrExpressionExtractor`, which keys off `SemanticModel.GetTypeInfo(reference).ConvertedType` for `[This]`-bearing delegates.
- **Direct invocation** of a `[This]`-bearing delegate from C# lowers to `handler.call(receiver, ...rest)`. `MethodKind.DelegateInvoke` + `[This]` on the first parameter triggers the `.call(...)` rewrite at `ExtractInvocation`, so the JS dispatcher sets `this` to the first argument before the `bindReceiver` trampoline forwards it. Without this rewrite, a direct invoke fires the delegate with `this === undefined` at the wrapper boundary.

## Slice D — cross-assembly + ADR

- **Cross-assembly**: `SymbolHelper.HasThis` already namespace-qualifies against `Metano.Annotations`, so a delegate declared in a referenced library propagates the attribute through `CompilationReference` symbols automatically. The new `This_CrossAssembly_DelegateReadFromReferencedLibrary` test pins the end-to-end shape (lib ships the delegate; consumer assigns a lambda; consumer's TS imports `bindReceiver` and emits the wrapped arrow).
- **Dart no-op**: already shipped via the #114 review fix — `IrToDartTypeMapper.BuildDartFunctionParameters` re-introduces `ThisType` as a positional parameter at index 0. Pinned by `DartBackendTests.ThisAttribute_DartBridge_ReintroducesReceiverAsFirstParameter`.
- **ADR-0016** records the design decision: `bindReceiver` runtime helper instead of a `function`-keyword + body-rewrite emission. The ADR documents the rejected alternative and the closure-collision bug it would carry, so a future revisit knows the trade space.

## Test plan

- [x] `dotnet run --project tests/Metano.Tests/` — **715/715 green** (710 → 715), 5 new tests in `ThisAttributeTranspileTests`:
  - `This_MethodGroupAssignment_WrapsReferenceInBindReceiver` — canonical `this.onClick = bindReceiver(this.handler)` shape
  - `This_StaticMethodGroupAssignment_WrapsReferenceInBindReceiver` — static-class-qualified path
  - `This_MethodGroupForPlainDelegate_NotWrapped` — regression guard, ordinary delegates keep the bare reference
  - `This_DirectInvocation_LowersThroughDotCall` — `.call(...)` rewrite for direct invokes
  - `This_CrossAssembly_DelegateReadFromReferencedLibrary` — `[This]` propagation through `CompilationReference`
- [ ] Reviewer: confirm the `WrapMethodGroupForThisDelegate` helper does not over-reach (only `IMethodSymbol` references with a delegate target carrying `[This]` get wrapped; ordinary identifier paths fall through unchanged)
- [ ] Reviewer: sanity-check the `.call(...)` rewrite handles both lambda-targets and method-group-targets at the receiver position (delegate held as a local / parameter)

Closes #113.

### DIFF
--- a/docs/adr/0016-this-attribute-bindreceiver.md
+++ b/docs/adr/0016-this-attribute-bindreceiver.md
@@ -1,0 +1,199 @@
+# ADR-0016 — `[This]` attribute via `bindReceiver` runtime helper
+
+**Status:** Accepted
+**Date:** 2026-04-24
+
+## Context
+
+JS/TS APIs frequently rely on the function's `this` binding rather
+than an explicit parameter. The canonical case is a DOM event
+handler:
+
+```ts
+element.onclick = function (event) {
+  this.innerHTML = "clicked";
+};
+```
+
+TypeScript expresses the contract through a synthetic first
+parameter on the function type:
+
+```ts
+type MouseEventListener = (this: Element, event: MouseEvent) => void;
+```
+
+C# has no native way to say "the first parameter IS the receiver."
+Without an explicit attribute, our DOM bindings either drop the
+receiver (losing the type information) or expose it as a regular
+positional argument that emits incorrectly at the dispatch boundary.
+
+Issue #113 introduced `[This]` to mark the first parameter of a
+delegate or inlinable method as the synthetic receiver. The
+attribute is parameter-only, applies strictly to the index-0 slot,
+and rejects `ref` / `out` / `in` / `params` modifiers (MS0018).
+
+## Decision
+
+The TypeScript backend lowers `[This]` through a thin runtime
+helper, **`bindReceiver`**, instead of synthesizing
+`function (this: T, ...) { ... }` blocks at every call site.
+
+### Helper
+
+```typescript
+export function bindReceiver<Receiver, Rest extends readonly unknown[], R>(
+  fn: (receiver: Receiver, ...rest: Rest) => R,
+): (this: Receiver, ...args: Rest) => R {
+  return function (this: Receiver, ...args: Rest): R {
+    return fn(this, ...args);
+  };
+}
+```
+
+The wrapper is a classic `function` so the dispatching runtime
+(DOM event loop, native API, manual `.call`) rebinds `this` on
+invocation. It then forwards that runtime `this` as the first
+positional argument to the wrapped arrow. The arrow itself stays
+lexically scoped — any `this` captured from the enclosing C# class
+flows through ordinary closure semantics, never aliased to the
+runtime receiver.
+
+### Generated shapes
+
+- **Delegate type declaration**: `(this: T, ...rest) => R` — keeps
+  the TS-idiomatic surface so consumers see the receiver shape on
+  the type line.
+- **Lambda assignment**: `bindReceiver((self, ...rest) => body)` —
+  the lambda's first parameter (the receiver, named by the user)
+  stays in the positional list; the helper feeds the dispatcher's
+  `this` into it. No body rewriting, no `function` keyword on the
+  generated lambda.
+- **Method-group assignment**: `bindReceiver(handler)` — the
+  reference is wrapped at the assignment site so the runtime can
+  trampoline the receiver through the named method.
+- **Direct invocation**: `handler.call(receiver, ...rest)` — when
+  C# code invokes a `[This]` delegate directly, the call is
+  rewritten to use `Function.prototype.call`, supplying the first
+  argument as the JS `this`.
+- **Dart**: no-op — the receiver re-introduces as a regular
+  positional parameter at index 0.
+
+### Rejected alternative: `function`-keyword emission with body rewrite
+
+The first iteration of Slice B emitted lambdas as
+`function (this: T, ...) { ... }` and rewrote every identifier
+referencing the receiver parameter into the keyword `this`. The
+shape was structurally correct for the receiver itself but silently
+collided with `this` references captured from the enclosing C#
+class:
+
+```csharp
+OnClick = self => self.InnerHtml = this.Name;  // `this` = Widget
+```
+
+emitted as
+
+```ts
+this.onClick = function (this: Element) {
+  return this.innerHtml = this.name;  // both `this` = Element. BUG.
+};
+```
+
+Each `function` keyword rebinds `this`, defeating the closure of
+the outer `Widget` instance. Patching this required emitting
+`const _self = this;` shims, a body walker, name-collision logic,
+and nested-scope bookkeeping — substantial machinery to defend
+against a footgun the runtime helper avoids by construction.
+
+`bindReceiver` keeps the lambda as a plain arrow, leaves the body
+untouched, and lets the V8/JSC closure semantics handle outer
+`this` for free.
+
+## Consequences
+
+- (+) DOM bindings emit ergonomic TypeScript (`element.onclick =
+      bindReceiver((self, e) => self.innerHTML = e)`) without
+      bespoke `function`-keyword printing or per-lambda body
+      rewriting.
+- (+) Outer-class `this` capture works through ordinary lexical
+      scope. The generated lambda body matches the C# author's
+      mental model.
+- (+) Method-group and lambda paths share a single mechanism — the
+      bridge wraps the reference / arrow in `bindReceiver` and the
+      ImportCollector auto-imports the helper from
+      `metano-runtime`. No new IR category for runtime
+      requirements.
+- (+) Cross-assembly works automatically — `SymbolHelper.HasThis`
+      already namespace-qualifies against `Metano.Annotations`, so
+      a delegate declared in a referenced library propagates the
+      attribute through `CompilationReference` symbols.
+- (+) Dart degrades cleanly — the receiver is re-introduced as a
+      positional parameter via `IrToDartTypeMapper.BuildDartFunctionParameters`.
+      No JS-only emission leaks into the Dart output.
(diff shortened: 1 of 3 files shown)


danfma:
@gemini-code-assist — addressed in `e7667b1` (force-pushed). The ADR's stack-trace caveat now mentions `x_google_ignoreList` explicitly as the standardized way to collapse the `bindReceiver` trampoline frame, with the link to Chrome's documentation and a note that listing `metano-runtime` under that key takes care of it without per-call annotations. Thread resolved.

danfma:
@copilot-pull-request-reviewer — all four threads addressed in `fdadfb7` (force-pushed):

- **Method-group reference unbound at `bindReceiver` call (P1)**: instance method-group references now `.bind(receiver)` before being handed to `bindReceiver`. The wrap detects `IrMemberAccess` whose target is not an `IrTypeReference` (i.e. instance receiver) and inserts a `.bind(receiver)` step. Static method groups skip the bind. The existing `This_MethodGroupAssignment_WrapsReferenceInBindReceiver` test now pins `bindReceiver(this.handler.bind(this))`.
- **`GenericNameSyntax` not wrapped**: `ExtractGenericName` now routes through `WrapMethodGroupForThisDelegate` on every emission path (instance / static / bare identifier), matching `ExtractIdentifierName`'s coverage. New test `This_GenericMethodGroupAssignment_WrapsReferenceInBindReceiver` pins `Handler<int>` flowing through the wrap with the `.bind(this)` step.
- **`.call(...)` rewrite under named arguments**: receiver argument is now identified by parameter name. `FindReceiverArgumentIndex` walks the `ArgumentSyntax` list looking for `NameColon` matching the receiver parameter name and falls back to position 0 only when every argument is positional. Mixed shapes (named arg targeting a non-receiver slot) bail out by returning `-1`, and the rewrite skips. New `This_DirectInvocation_NamedReceiverArgument_LowersThroughDotCall` pins `handler(arg: "fired", self: target)` lowering to `handler.call(target, "fired")`.
- **`.call` rewrite over low-precedence receiver expressions**: new `IsSafeForCallRewrite` guard skips the rewrite when the invoked expression unwraps to `ConditionalExpressionSyntax`, `BinaryExpressionSyntax`, `AssignmentExpressionSyntax`, or `AwaitExpressionSyntax`. Without parens around the receiver, `(a ?? b)(args)` would have printed as `a ?? b.call(args)` and changed the parse. The fallback path emits the plain delegate invocation — semantically off (matches the legacy pre-rewrite behavior), but no longer malformed. Tracked as a follow-up to introduce IR-level parenthesization.

Tests: **717/717 green** (715 → 717, 2 new). All 4 review threads resolved.

PR #125, opened 2026-04-25
fix(ts): emit static members declared on records

### LINKED ISSUE #119: Static members inside records are missing from the generated TS
## Motivation

Records commonly expose cached singletons or factory helpers as static members:

```csharp
[Transpile]
public sealed record Counter(int Count)
{
    public static readonly Counter Zero = new(0);

    public Counter Increment() => this with { Count = Count + 1 };
}
```

Expected TypeScript output keeps \`Zero\` as a static readonly member on the emitted class (or as a top-level \`export const Zero\` when the record maps to \`[PlainObject]\`):

```ts
export class Counter {
  static readonly Zero = new Counter(0);
  constructor(public readonly count: number) {}
  increment(): Counter { return new Counter(this.count + 1); }
}
```

Today the static member is silently dropped: the generated \`counter.ts\` contains the positional fields, the \`increment\` method, and nothing else. Call sites that reference \`Counter.Zero\` either fail to compile or resolve to \`undefined\`.

## Repro

```csharp
public sealed record Counter(int Count)
{
    public static readonly Counter Zero = new(0);

    public Counter Increment() => this with { Count = Count + 1 };
}
```

Run the existing sample / test pipeline and inspect the emitted \`counter.ts\`. The \`Zero\` declaration is absent.

## Likely cause

\`IrClassExtractor\` walks \`GetMembers()\` for record types but the static-member branches (fields with initializer, static properties, static methods) may be filtered out by the record-specific path that focuses on positional properties. A quick \`grep\` around \`IrClassExtractor.cs\` for the record branch and \`IsStatic\` filtering should confirm.

## Acceptance

- [ ] Static readonly fields with initializers on records emit as \`static readonly name = <init>\` on the generated TS class.
- [ ] Static properties and static methods on records emit the same way they do on plain classes.
- [ ] \`[Inline]\` static members on records still honor the call-site substitution contract (no declaration emitted).
- [ ] Tests:
  - Static readonly field with a record-typed initializer (e.g. \`Zero\`).
  - Static method returning a record.
  - Static property on a record with \`[Name]\` override.
  - \`[PlainObject]\` record: static members promoted to top-level \`export const\` in the record's module file.

danfma:
## Summary

Closes #119. Static **fields** declared on a record type were silently dropped from the generated TypeScript output:

```csharp
[Transpile]
public sealed record Counter(int Count)
{
    public static readonly Counter Zero = new(0);
}
```

emitted a `Counter` class that had the positional fields, the `with`/`equals`/`hashCode` triplet, but no `Zero` declaration anywhere. Static methods and static properties on records were already correct.

## Fix

**Class bridge** (`IrToTsClassBridge.BuildField`)
- Propagates `field.IsStatic` into `TsFieldMember.Static`. The TS printer was already aware of the slot; the bridge just never set it. Static fields keep their explicit initializer; the default-initializer fallback (`= 0`, `= false`, etc.) only applies to instance fields the constructor would otherwise leave undefined.

**PlainObject bridge** (`IrToTsPlainObjectBridge.Convert`)
- Static fields with an initializer now emit as a top-level `export const Name = <init>;` in the record's module file (PlainObject records have no class wrapper). Honors `[Name]` overrides via the existing naming policy and the standard expression bridge for the initializer.
- Static methods on PlainObject records stay deferred — the shape of "static utility on a wire-shape record" deserves its own ADR-style decision and is tracked as a follow-up. Instance fields on a PlainObject record were already shape-incompatible (interfaces don't carry storage); no change there.

## Test plan

- [x] `dotnet run --project tests/Metano.Tests/` — **721/721 green** (717 → 721), 4 new tests in `RecordStaticMembersTests`:
  - `RecordWithStaticReadonlyField_EmitsFieldOnGeneratedClass` — canonical `static readonly zero = new Counter(0)` shape
  - `RecordWithStaticMethod_EmitsMethodOnGeneratedClass` — regression guard
  - `RecordWithStaticProperty_EmitsPropertyOnGeneratedClass` — regression guard
  - `PlainObjectRecordWithStaticField_EmitsTopLevelExportConst` — PlainObject path

Closes #119.

### DIFF
--- a/src/Metano.Compiler.TypeScript/Bridge/IrToTsClassBridge.cs
+++ b/src/Metano.Compiler.TypeScript/Bridge/IrToTsClassBridge.cs
@@ -808,13 +808,19 @@ internal static class IrToTsClassBridge
             field.Initializer is not null && !field.IsCapturedByCtor
                 ? IrToTsExpressionBridge.Map(field.Initializer, bclRegistry)
                 : null;
-        initializer ??= ComputeDefaultInitializer(field.Type);
+        // Static fields with an explicit initializer are the canonical
+        // shape (`static readonly Zero = new Counter(0)`); the
+        // default-initializer fallback only makes sense for instance
+        // fields that the constructor would otherwise leave undefined.
+        if (!field.IsStatic)
+            initializer ??= ComputeDefaultInitializer(field.Type);
 
         return new TsFieldMember(
             name,
             tsType,
             initializer,
             field.IsReadonly,
+            Static: field.IsStatic,
             Accessibility: MapAccessibility(field.Visibility)
         );
     }
--- a/src/Metano.Compiler.TypeScript/Bridge/IrToTsClassBridge.cs
+++ b/src/Metano.Compiler.TypeScript/Bridge/IrToTsClassBridge.cs
@@ -808,12 +808,14 @@ internal static class IrToTsClassBridge
             field.Initializer is not null && !field.IsCapturedByCtor
                 ? IrToTsExpressionBridge.Map(field.Initializer, bclRegistry)
                 : null;
-        // Static fields with an explicit initializer are the canonical
-        // shape (`static readonly Zero = new Counter(0)`); the
-        // default-initializer fallback only makes sense for instance
-        // fields that the constructor would otherwise leave undefined.
-        if (!field.IsStatic)
-            initializer ??= ComputeDefaultInitializer(field.Type);
+        // Mirror C#'s zero-initialization story for both instance and
+        // static fields: a C# field without an explicit initializer
+        // still reads as the type's default at runtime (`0`,
+        // `false`, `null`, …). TypeScript leaves a no-initializer
+        // field as `undefined`, which would silently shift behavior
+        // — fall back to the matching default so the generated
+        // class keeps the C# semantics.
+        initializer ??= ComputeDefaultInitializer(field.Type);
 
         return new TsFieldMember(
             name,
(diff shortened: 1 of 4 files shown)


danfma:
@gemini-code-assist @copilot-pull-request-reviewer — all 8 threads addressed across `a28dbc3` and `53bb2ab`.

**Gemini findings**:
- **Static fields lost C# zero-init** → `BuildField` now applies `ComputeDefaultInitializer` for both instance and static fields, restoring the C# default-value semantics.
- **`ToTypeName` for PlainObject field name** → switched to `ToInterfaceMemberName` so the export uses the same camelCased policy as every other field.
- **Hardcoded `Const: true`** → derives the flag from `field.IsReadOnly`. Mutable statics lower to `export let`. New `PlainObjectRecordWithMutableStaticField_EmitsTopLevelExportLet` pins the behavior.
- **Test expectation casing** → flipped to `export const zero` to match the camelCased policy.

**Copilot findings** (PlainObject path):
- **Reserved-word override produced invalid TS** → new `TypeScriptNaming.EscapeIfReserved` appends a trailing underscore at top-level declaration sites where the language rejects reserved identifiers (`export const delete = ...` is illegal). Used in `EmitStaticField`. New `PlainObjectRecordWithReservedNameOverride_EscapesIdentifier` pins `[Name("delete")]` → `export const delete_`.
- **Default initializer for static fields** → covered by the Gemini fix above; both reviewers landed on the same root cause.
- **Visibility filter leaked private/protected** → restricted to `IrVisibility.Public`; non-public statics no longer surface as top-level exports. New `PlainObjectRecordWithPrivateStaticField_DoesNotExport` pins the filter.
- **`Const: true` ignored mutability** → covered by the Gemini fix above.

Tests: **724/724 green** (721 → 724, 3 net new tests across the two fix commits). All 8 review threads resolved.

PR #126, opened 2026-04-25
fix(core): lower a?.b = c instead of raising Unsupported

### LINKED ISSUE #120: Null-conditional assignment: lower a?.b = c instead of raising Unsupported
## Motivation

Sample code pulled from \`samples/SampleCounterWithDomView/Views/CounterView.cs\`:

```csharp
_text?.InnerHtml = counter.ToString();
```

Today the extractor surfaces:

\`\`\`
unsupported IR expression ConditionalAccess(SimpleAssignmentExpression)
\`\`\`

which leaves the statement unrendered in the generated TypeScript. The message originates in \`IrExpressionExtractor.LowerConditionalBinding\` — the switch only handles \`MemberBindingExpressionSyntax\`, \`InvocationExpressionSyntax\`, and nested \`ConditionalAccessExpressionSyntax\`, and falls through to \`IrUnsupportedExpression\` when the right-hand side is an assignment.

## Proposal

Lower \`a?.b = c\` to a guarded assignment. TypeScript does not support optional-chained assignment directly, so emit:

```ts
if (a !== null && a !== undefined) {
  a.b = c;
}
```

or, for the common single-dereference form and when the receiver expression is side-effect free, an inline conditional:

```ts
a !== null && a !== undefined && (a.b = c);
```

The guarded statement form is preferred for readability; the inline conditional is only used when the construct sits inside an expression context (e.g., inside another assignment).

## Semantics

- **Simple property LHS.** \`a?.b = c\` → guarded statement as above.
- **Chained access.** \`a?.b?.c = d\` → \`if (a !== null && a !== undefined && a.b !== null && a.b !== undefined) { a.b.c = d; }\`. Avoid repeating evaluation of \`a.b\` when it is an expression with side effects — introduce a synthetic const.
- **Compound assignment.** \`a?.b += c\` → \`if (a !== null && a !== undefined) { a.b += c; }\`.
- **Element access.** \`a?[i] = v\` → \`if (a !== null && a !== undefined) { a[i] = v; }\`.
- **Side-effect receiver.** When \`a\` is a method call, hoist to a local so the guard and the assignment share the same evaluation.

## IR shape

Introduce an \`IrConditionalAssignment\` (receiver, assignment-target lens, value, compound-op?) that bridges render as the guarded statement. Extraction maps \`ConditionalAccessExpressionSyntax\` whose \`WhenNotNull\` is an assignment (possibly a chain ending in an assignment) to this shape.

## Open questions

- **Strict-null vs truthy check.** Standard metano lowering of \`?.\` uses \`!== null && !== undefined\` to avoid falsy-coercion traps. Keep that convention.
- **Return value.** In C# an assignment expression yields the assigned value; in the null-conditional form the expression type is \`void\`. Lower to a statement when the surrounding context is a statement; lower to a conditional-yielding expression (\`a != null ? (a.b = c, c) : undefined\`) when embedded in an expression context.
- **Dart target.** Mirror with \`a?.b = c\`, which Dart supports natively. The cross-target IR should still carry the shape explicitly.

## Acceptance

- [ ] \`IrConditionalAssignment\` (or equivalent) added to IR.
- [ ] Extraction covers \`ConditionalAccessExpressionSyntax\` whose \`WhenNotNull\` leads to an \`AssignmentExpressionSyntax\` (simple and compound).
- [ ] TS bridge emits the guarded statement (or expression form when required).
- [ ] Receivers with side effects hoist to a synthetic local.
- [ ] Dart bridge emits the native \`?.\` assignment.
- [ ] Tests:
  - Simple \`a?.b = c\`.
  - Chained \`a?.b?.c = d\`.
  - Compound \`a?.b += c\`.
  - Element access \`a?[i] = v\`.
  - Receiver with side effects (method call).
  - Regression: \`samples/SampleCounterWithDomView/Views/CounterView.cs\` transpiles end-to-end.

danfma:
## Summary

Closes #120. C# null-conditional assignment (`a?.b = c`) was extracted as `IrUnsupportedExpression`, leaving the statement out of the generated TypeScript. The new branch lowers the shape to an inline short-circuit so the null-guard is preserved without a body-walker rewrite.

```csharp
_text?.InnerHtml = counter.ToString();
```

now emits

```ts
this._text != null && (this._text.innerHtml = counter.toString());
```

Loose equality (per ADR-0014) covers `null` and `undefined`. The receiver is referenced twice in the emitted TS, which is safe for the dominant simple-receiver shape (identifier / member access). Side-effecting receivers would re-evaluate — documented in the comment as a known edge case until the IR grows a let-binding shape that can host the temp.

Compound assignment forms (`a?.b += c`, `*=`, etc.) flow through the same path; the operator carried in `AssignmentExpressionSyntax.Kind()` maps via the existing `MapAssignmentOp` helper.

## Test plan

- [x] `dotnet run --project tests/Metano.Tests/` — **727/727 green** (724 → 727), 3 new tests in `NullConditionalAssignmentTests`:
  - `NullConditionalPropertyAssignment_LowersThroughShortCircuit` pins the canonical `this._text != null && (this._text.innerHtml = …)` shape
  - `NullConditionalAssignment_OnLocal_LowersWithoutThis` covers local receivers (no implicit `this.`)
  - `NullConditionalCompoundAssignment_LowersWithMatchingOperator` pins `+=` operator preservation
- [ ] Reviewer: confirm the receiver-duplication trade-off is acceptable for the simple-receiver case (identifier / member access)

Closes #120.

### DIFF
--- a/src/Metano.Compiler.Dart/Dart/IrBodyPrinter.cs
+++ b/src/Metano.Compiler.Dart/Dart/IrBodyPrinter.cs
@@ -783,6 +783,7 @@ public static class IrBodyPrinter
             IrBinaryOp.BitwiseXorAssign => "^=",
             IrBinaryOp.LeftShiftAssign => "<<=",
             IrBinaryOp.RightShiftAssign => ">>=",
+            IrBinaryOp.UnsignedRightShiftAssign => ">>>=",
             IrBinaryOp.NullCoalescingAssign => "??=",
             _ => "/* ?? */",
         };
--- a/src/Metano.Compiler.TypeScript/Bridge/IrToTsExpressionBridge.cs
+++ b/src/Metano.Compiler.TypeScript/Bridge/IrToTsExpressionBridge.cs
@@ -183,6 +183,7 @@ public static class IrToTsExpressionBridge
                 or IrBinaryOp.BitwiseXorAssign
                 or IrBinaryOp.LeftShiftAssign
                 or IrBinaryOp.RightShiftAssign
+                or IrBinaryOp.UnsignedRightShiftAssign
                 or IrBinaryOp.NullCoalescingAssign
                 or IrBinaryOp.NullCoalescing;
 
@@ -921,6 +922,7 @@ public static class IrToTsExpressionBridge
             IrBinaryOp.BitwiseXorAssign => "^=",
             IrBinaryOp.LeftShiftAssign => "<<=",
             IrBinaryOp.RightShiftAssign => ">>=",
+            IrBinaryOp.UnsignedRightShiftAssign => ">>>=",
             IrBinaryOp.NullCoalescingAssign => "??=",
             _ => "?",
         };
--- a/src/Metano.Compiler/Extraction/IrExpressionExtractor.cs
+++ b/src/Metano.Compiler/Extraction/IrExpressionExtractor.cs
@@ -158,6 +158,35 @@ public sealed class IrExpressionExtractor
 
                 return LowerConditionalBinding(head, nested.WhenNotNull);
 
+            case AssignmentExpressionSyntax assign
+                when assign.Left is MemberBindingExpressionSyntax memberBinding:
+                // `a?.b = c` lowers to `a != null && (a.b = c)`.
+                // TypeScript has no optional-chained assignment, so
+                // the null-guard becomes an inline short-circuit
+                // expression. The receiver is evaluated only once
+                // when it is a simple identifier / member access (the
+                // overwhelming majority of real call sites); a
+                // side-effecting receiver would re-evaluate, which is
+                // documented as a known edge case until the IR
+                // grows a let-binding shape that can host the temp.
+                var assignedValue = Extract(assign.Right);
+                var memberName = memberBinding.Name.Identifier.ValueText;
+                var memberOrigin = BuildOrigin(_semantic.GetSymbolInfo(memberBinding).Symbol);
+                var memberWrite = new IrBinaryExpression(
+                    new IrMemberAccess(receiver, memberName, memberOrigin),
+                    MapAssignmentOp(assign.Kind()),
+                    assignedValue
+                );
+                return new IrBinaryExpression(
+                    new IrBinaryExpression(
+                        receiver,
+                        IrBinaryOp.NotEqual,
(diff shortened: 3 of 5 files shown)


danfma:
@gemini-code-assist @copilot-pull-request-reviewer — all 6 threads addressed in `5d239b8`. The previous inline lowering moved into a focused `TryLowerNullConditionalAssignment` helper so the diagnostic / fall-through path stays clean.

**Gemini findings**:
- **Receiver duplication**: new `IsSimpleReceiver` guard restricts the rewrite to identifier / `this` / `base` / type-reference / chains of member-access nodes. Method calls, indexers, and complex expressions fall through to the unsupported path until the IR grows a let-binding shape.
- **Nested member access / indexer**: new `RebindAssignmentTarget` helper walks the LHS recursively, swapping the inner `MemberBindingExpression` for `IrMemberAccess(receiver, name)` and threading additional member / element accesses on top. `a?.b.c = d` and `a?.items[0] = e` now lower through the same short-circuit. New `NullConditionalNestedMemberAssignment_LowersThroughShortCircuit` test pins the shape.
- **Event accessor leak**: bail out when the LHS resolves to an `IEventSymbol`, so the existing `event += handler` runtime path (`delegateAdd` / `delegateRemove`) keeps its dispatch instead of being silently flattened to a plain compound assignment.

**Copilot findings**:
- **`>>>=` operator missing**: added `UnsignedRightShiftAssign` to `IrBinaryOp`, wired the `MapAssignmentOp` mapping, the TypeScript bridge token (`>>>=`) + precedence list, and the Dart printer (Dart uses the same token). `a?.b >>>= c` no longer falls back to `IrBinaryOp.Assign`.
- **Dart compatibility**: gated the lowering to the TypeScript target (or null target for tests). Dart's `&&` requires a boolean RHS, which an assignment expression does not satisfy. Other backends fall through to the unsupported path; a future Dart-friendly statement-level rewrite (`if (a != null) a.b = c;`) can ship under its own slice.
- **Stale "evaluated only once" comment**: rewrote the docstring to acknowledge the receiver is referenced twice in the emitted code and to enumerate the IR shapes that are safe under that constraint (covered by `IsSimpleReceiver`).

Tests: **728/728 green** (727 → 728). All 6 review threads resolved.

PR #127, opened 2026-04-25
feat(core): lower handler?.Invoke(args) to optional call

### LINKED ISSUE #121: Null-conditional delegate invocation: lower handler?.Invoke(...) to handler?.(...)
## Motivation

The canonical C# null-safe event pattern uses \`?.Invoke(...)\`:

```csharp
button.onclick = (_: MouseEvent) => this.onButtonClick?.Invoke();
```

TypeScript has the equivalent optional-call syntax:

```ts
button.onclick = (_: MouseEvent) => { this.onButtonClick?.(); };
```

Today \`IrExpressionExtractor.LowerConditionalBinding\` handles the invocation branch only when the invocation's target is a \`MemberBindingExpressionSyntax\` (\`a?.Method()\` against a member name). The delegate-invoke shape — \`a?.Invoke(...)\` where \`Invoke\` is the synthesized delegate method — lowers as a regular call, erasing the null guard, or in some combinations surfaces as \`Unsupported\`.

## Proposal

Detect the invocation whose symbol points to a delegate's \`Invoke\` method and emit an optional call expression (\`IrOptionalCall\`) that the bridges render as:

- TypeScript: \`handler?.(...args)\`
- Dart: \`handler?.call(...args)\`

Also cover the non-conditional \`handler.Invoke(arg)\` form — emit \`handler(arg)\` (already a common ergonomic ask for delegate calls).

## Semantics

- **\`handler?.Invoke(args)\`** → \`handler?.(args)\` in TS.
- **\`handler.Invoke(args)\`** → \`handler(args)\` in TS (omits the \`.Invoke\` indirection that has no runtime counterpart in JS).
- **\`handler?.Invoke()\` inside an expression** → \`handler?.()\`, guaranteed to short-circuit.
- **Event-backed delegates.** The \`_text\$add\` / \`_text\$remove\` helpers emitted via \`delegateAdd\` / \`delegateRemove\` already exist; this path leaves those accessors untouched and only simplifies consumer-side invocations.

## IR shape

- Add an \`IrOptionalCall(IrExpression receiver, IReadOnlyList<IrArgument> args, IrMemberOrigin? origin)\` (or reuse \`IrCallExpression\` with an \`IsOptional\` flag).
- Extraction detects the invocation target symbol is a method named \`Invoke\` on an \`INamedTypeSymbol\` with \`TypeKind.Delegate\`. When the outer context is a \`ConditionalAccessExpressionSyntax\`, emit the optional form; otherwise emit a plain call on the delegate receiver.

## Open questions

- **Compound delegate chains.** \`a?.b?.Invoke(c)\` composes with the member optional-chain already in place. Verify the recursive \`LowerConditionalBinding\` still threads the call through after the nested receiver resolves.
- **Static delegate fields.** \`Handler?.Invoke()\` where \`Handler\` is \`static\` resolves to a bare identifier at the call site; the optional-call form must not accidentally promote it to \`this.handler\`.
- **Dart target.** \`handler?.call(...)\` is the native shape; make sure the IR carries enough information for the Dart bridge.

## Acceptance

- [ ] IR carries an \`IrOptionalCall\` (or equivalent flag on \`IrCallExpression\`).
- [ ] Extraction recognizes \`Invoke\` on delegate types and emits the optional form when the context is \`ConditionalAccessExpressionSyntax\`.
- [ ] Non-conditional \`handler.Invoke(...)\` lowers to \`handler(...)\`.
- [ ] TypeScript bridge emits \`receiver?.(...)\` / \`receiver(...)\` according to the IR flag.
- [ ] Dart bridge emits \`receiver?.call(...)\` / \`receiver.call(...)\`.
- [ ] Tests:
  - \`handler?.Invoke()\` → \`handler?.()\`.
  - \`handler?.Invoke(a, b)\` → \`handler?.(a, b)\`.
  - \`handler.Invoke(a)\` → \`handler(a)\`.
  - Static delegate field lowering (no implicit \`this.\`).
  - Chained \`this.onClick?.Invoke()\` inside a lambda (exercises the sample shape).
  - Regression: \`samples/SampleCounterWithDomView\` renders the event lambda without \`Unsupported\` markers.

danfma:
## Summary

- Lower C# `handler?.Invoke(args)` to TypeScript `handler?.(args)` and Dart `handler?.call(args)`.
- Drop the `.Invoke` indirection on non-conditional `handler.Invoke(args)` (becomes `handler(args)` on both targets) — the runtime delegate is the function.
- `[This]`-bearing delegates fall through both shortcuts so the existing `delegate.call(receiver, ...)` rewrite keeps producing the JS `this`-rebind shape.

## Mechanism

- `IrCallExpression` gains an `IsOptional` flag.
- `IrExpressionExtractor` detects `MethodKind.DelegateInvoke` at two sites (conditional binding via `MemberBindingExpressionSyntax` and direct invocation via `MemberAccessExpressionSyntax`), drops the binding, and sets the flag.
- TS printer: emits `?.` between callee and argument list when `Optional`.
- Dart printer: emits `?.call` when `IsOptional`.

## Test plan

- [x] 11 new tests in `NullConditionalInvokeTests.cs` covering: TS no-args, TS with-args, TS direct, TS static field, TS inside lambda, TS generic `Func<T,R>`, TS `[This]`-bearing (keeps `.call` rewrite), TS event `+=` (unaffected), Dart conditional, Dart direct.
- [x] Full suite: `dotnet run --project tests/Metano.Tests/` → 738/738 pass.
- [x] csharpier-format clean.

Closes #121

### DIFF
--- a/src/Metano.Compiler.Dart/Dart/IrBodyPrinter.cs
+++ b/src/Metano.Compiler.Dart/Dart/IrBodyPrinter.cs
@@ -252,7 +252,7 @@ public static class IrBodyPrinter
                 sb.Append(']');
                 break;
             case IrCallExpression call:
-                PrintExpression(sb, call.Target);
+                PrintCallTarget(sb, call.Target);
                 if (call.IsOptional)
                     sb.Append("?.call");
                 sb.Append('(');
@@ -360,6 +360,35 @@ public static class IrBodyPrinter
                 break;
         }
     }
+
+    /// <summary>
+    /// Print a call target, wrapping it in parentheses when its node
+    /// kind has lower precedence than the call. Required because the
+    /// IR may surface low-precedence shapes as the call target (e.g.
+    /// <c>(a ?? b)?.Invoke()</c> lowers to a call whose target is a
+    /// binary expression — printed bare it would parse as
+    /// <c>a ?? b()</c>).
+    /// </summary>
+    private static void PrintCallTarget(StringBuilder sb, IrExpression target)
+    {
+        if (CallTargetRequiresParens(target))
+        {
+            sb.Append('(');
+            PrintExpression(sb, target);
+            sb.Append(')');
+        }
+        else
+        {
+            PrintExpression(sb, target);
+        }
+    }
+
+    private static bool CallTargetRequiresParens(IrExpression expression) =>
+        expression
+            is IrBinaryExpression
+                or IrConditionalExpression
+                or IrAwaitExpression
+                or IrLambdaExpression;
 
     /// <summary>
     /// Dart renders <c>x is Foo</c> natively. A type-with-designator (<c>x is Foo f</c>)
--- a/src/Metano.Compiler.Dart/Dart/IrBodyPrinter.cs
+++ b/src/Metano.Compiler.Dart/Dart/IrBodyPrinter.cs
@@ -253,6 +253,8 @@ public static class IrBodyPrinter
                 break;
             case IrCallExpression call:
                 PrintExpression(sb, call.Target);
+                if (call.IsOptional)
+                    sb.Append("?.call");
                 sb.Append('(');
                 PrintArgs(sb, call.Arguments);
                 sb.Append(')');
--- a/src/Metano.Compiler.TypeScript/Bridge/IrToTsExpressionBridge.cs
+++ b/src/Metano.Compiler.TypeScript/Bridge/IrToTsExpressionBridge.cs
@@ -313,7 +313,7 @@ public static class IrToTsExpressionBridge
                 return mapped;
         }
 
-        return new TsCallExpression(Map(call.Target, bclRegistry), loweredArgs);
+        return new TsCallExpression(Map(call.Target, bclRegistry), loweredArgs, call.IsOptional);
     }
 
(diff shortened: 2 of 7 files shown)


danfma:
@copilot-pull-request-reviewer thanks for catching the parenthesization gap. Fixed in da39829.

Both findings collapse into the same problem: the new Invoke shortcuts can surface low-precedence shapes as the call callee/target. The fix lives in the printers rather than the bridge so it covers both targets uniformly:

- `Printer.PrintCallCallee` wraps the callee in parens when it's `TsBinaryExpression`, `TsConditionalExpression`, `TsAwaitExpression`, or `TsArrowFunction`.
- `IrBodyPrinter.PrintCallTarget` does the same against the IR shapes (`IrBinaryExpression`, `IrConditionalExpression`, `IrAwaitExpression`, `IrLambdaExpression`) for Dart.

Three regression tests added:
- `(handler ?? fallback)?.Invoke()` → `(this.handler ?? this.fallback)?.()`
- `(handler ?? fallback).Invoke()` → `(this.handler ?? this.fallback)()`
- chained `a?.b?.c?.Invoke()` → `a?.b?.c?.()` — each short-circuit preserved through the `IrOptionalChain` recursion

741/741 tests pass.

PR #128, opened 2026-04-25
feat(ts): emit extends clause for interface inheritance

### LINKED ISSUE #117: Interface inheritance: emit extends chain for IA : IB
## Motivation

Given a C# interface hierarchy:

```csharp
[Transpile]
public interface ICloseable
{
    void Close();
}

[Transpile]
public interface IStream : ICloseable
{
    int Read();
}
```

The expected TypeScript output is:

```ts
export interface ICloseable {
  close(): void;
}

export interface IStream extends ICloseable {
  read(): number;
}
```

\`IrInterfaceExtractor.cs:47-49\` already captures the base interfaces (\`type.Interfaces\`) into the IR, but the TypeScript bridge does not currently propagate that chain into the emitted \`extends\` clause for every shape, so consumers see duplicated member signatures or orphan interfaces.

## Proposal

Thread the IR \`BaseInterfaces\` list through \`IrToTsInterfaceBridge\` (and the interface-facing paths in \`IrToTsClassBridge\`, \`IrToTsPlainObjectBridge\`, and \`IrToTsRecordSynthesisBridge\`) so every generated \`export interface\` carries \`extends A, B, C\` when the source declares base interfaces.

Naming honors existing policy:
- \`--strip-interface-prefix\` (see #92 follow-up) rewrites both the declaration name and the \`extends\` entries consistently.
- \`[Name(...)]\` overrides apply to the base reference too.
- Cross-package base interfaces get an \`import type { IBase } from "pkg/path"\` entry via the existing import-collection pass.

## Semantics

- **Single base.** \`interface IA : IB\` → \`interface IA extends IB\`.
- **Multiple bases.** \`interface IA : IB, IC\` → \`interface IA extends IB, IC\`.
- **Generic bases.** \`interface IA<T> : IEnumerable<T>\` → \`interface IA<T> extends Iterable<T>\` using the current BCL mapping rules.
- **\`[PlainObject]\` records treated as interfaces** inherit the base-interface chain on the emitted TS interface. The constructor-side \`create(...)\` helper keeps the current shape.

## Open questions

- **Mixed transpilable and non-transpilable bases.** When a base interface is \`[NoEmit]\` / \`[External]\`, treat it as an ambient type — still emit it in \`extends\` but without a matching import. Existing behavior for class base types should apply symmetrically.
- **Diamond inheritance.** TypeScript interfaces merge compatible members; no extra work needed, but verify that the import collector deduplicates shared ancestors.

## Acceptance

- [ ] \`IrToTsInterfaceBridge\` emits \`extends A, B\` when \`BaseInterfaces\` is non-empty.
- [ ] \`[PlainObject]\` + \`IrToTsRecordSynthesisBridge\` interface outputs honor the chain.
- [ ] Name overrides (\`[Name]\`) and \`--strip-interface-prefix\` apply to base entries.
- [ ] Import collector adds cross-package \`import type\` entries for non-local bases.
- [ ] Tests:
  - Single base.
  - Multiple bases.
  - Generic base with a BCL mapping (\`IEnumerable<T>\` → \`Iterable<T>\`).
  - Cross-package base (via \`TranspileWithLibrary\`).
  - \`[Name]\` override on the base.

danfma:
## Summary

C# `interface IA : IB, IC` now lowers to TypeScript `interface IA extends IB, IC`. The IR already captured the base chain (`IrInterfaceDeclaration.BaseInterfaces`); this PR threads it through the TS bridge, AST, printer, and import collector.

## Mechanism

- `TsInterface` AST: new `Extends: IReadOnlyList<TsType>?` field.
- `IrToTsInterfaceBridge.BuildExtends`: mirrors `IrToTsClassBridge.BuildImplements` — maps each base via `IrToTsTypeMapper`, filters out `IrNamedTypeRef { IsTranspilable: false }` (BCL bases without `[ExportFromBcl]`).
- `IrToTsPlainObjectBridge`: passes `BuildImplements(ir)` to `TsInterface.Extends` so `[PlainObject]` records that implement interfaces emit with the chain.
- `Printer.PrintInterface`: emits ` extends A, B` after type params, before the block.
- `ImportCollector.CollectFromTopLevel` (`TsInterface` arm): walks `iface.Extends` so cross-package bases generate the right `import type { IBase } from "pkg/path"` entry and `package.json` dependency.

## Side effect: cross-package member types on interfaces

`TypeTransformer` now wires `Context.OriginResolver` into `IrInterfaceExtractor.Extract`. The extractor previously got `null`, which silently dropped origin metadata for cross-package types referenced by interface members (not just bases). They were emitted as bare names with no import. The test `CrossPackageMemberType_OnInterface_AddsImport` locks this in. Same-package output is byte-identical because origin resolution returns null for local types.

## Test plan

- [x] 10 new tests in `InterfaceInheritanceTests.cs`: single base, multiple bases, generic base, BCL generic base (`IReadOnlyCollection<T>` → `Iterable<T>`), `[Name]` override on base, `--strip-interface-prefix`, cross-package base import, non-transpilable filter (`IDisposable`), `[PlainObject]` record extends interface, cross-package member type.
- [x] Full suite: 750/750 pass.
- [x] csharpier-format clean.

## Out of scope

- Dart target — issue is TS-only. The Dart bridge already emits `implements` from `IrInterfaceDeclaration.BaseInterfaces` directly; no changes needed.

Closes #117

### DIFF
--- a/src/Metano.Compiler.TypeScript/Bridge/IrToTsClassBridge.cs
+++ b/src/Metano.Compiler.TypeScript/Bridge/IrToTsClassBridge.cs
@@ -39,16 +39,32 @@ internal static class IrToTsClassBridge
     /// interface that doesn't get emitted alongside the class. Returns
     /// <c>null</c> when nothing remains so the printer elides the clause.
     /// </summary>
-    public static IReadOnlyList<TsType>? BuildImplements(IrClassDeclaration ir)
+    public static IReadOnlyList<TsType>? BuildImplements(IrClassDeclaration ir) =>
+        ConvertHeritageList(ir.Interfaces);
+
+    /// <summary>
+    /// Lowers a list of IR base / interface references into TS heritage
+    /// entries. Drops named refs flagged <c>IsTranspilable: false</c>
+    /// (e.g. <c>[NoEmit]</c> or unmapped BCL interfaces) and rewrites
+    /// the array shorthand <c>T[]</c> (produced when a base maps
+    /// through <see cref="IrArrayTypeRef"/> for collection-like BCL
+    /// types) into <c>Array&lt;T&gt;</c>: TypeScript rejects
+    /// <c>extends T[]</c> / <c>implements T[]</c> in heritage clauses,
+    /// but accepts the named <c>Array</c> form.
+    /// </summary>
+    public static IReadOnlyList<TsType>? ConvertHeritageList(IReadOnlyList<IrTypeRef>? references)
     {
-        if (ir.Interfaces is not { Count: > 0 } ifaces)
+        if (references is not { Count: > 0 } refs)
             return null;
         var result = new List<TsType>();
-        foreach (var iface in ifaces)
+        foreach (var reference in refs)
         {
-            if (iface is IrNamedTypeRef { Semantics.IsTranspilable: false })
+            if (reference is IrNamedTypeRef { Semantics.IsTranspilable: false })
                 continue;
-            result.Add(IrToTsTypeMapper.Map(iface));
+            var mapped = IrToTsTypeMapper.Map(reference);
+            result.Add(
+                mapped is TsArrayType array ? new TsNamedType("Array", [array.ElementType]) : mapped
+            );
         }
         return result.Count > 0 ? result : null;
     }
--- a/src/Metano.Compiler.TypeScript/Bridge/IrToTsInterfaceBridge.cs
+++ b/src/Metano.Compiler.TypeScript/Bridge/IrToTsInterfaceBridge.cs
@@ -45,15 +45,39 @@ public static class IrToTsInterfaceBridge
         // consequently the file name + import collector).
         var tsName = nameOverride ?? IrToTsNamingPolicy.ToTypeName(ir.Name, ir.Attributes);
         var typeParams = ConvertTypeParameters(ir.TypeParameters);
+        var extends = BuildExtends(ir.BaseInterfaces);
 
         statements.Add(
             new TsInterface(
                 tsName,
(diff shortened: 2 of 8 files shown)


danfma:
@gemini-code-assist @chatgpt-codex-connector @copilot-pull-request-reviewer thanks for the review. All four findings addressed in 3f85a95.

- **DRY (Gemini)** — extracted `ConvertHeritageList` on `IrToTsClassBridge`. `BuildImplements` and `BuildExtends` both delegate to it; same filter + same array-shorthand normalization apply to class `implements` and interface `extends` clauses.
- **Array shorthand (Codex P1)** — genuine bug. `IEnumerable<T>` / `IList<T>` etc. map through `IrArrayTypeRef`, which the printer emits as `T[]`. TS rejects that shape in heritage clauses. The shared helper now rewrites `TsArrayType` → `TsNamedType("Array", element)` so `interface IRows<T> : IEnumerable<T>` emits `extends Array<T>`. Locked in by `CollectionLikeBase_NormalizesArrayShorthandToNamedArray`.
- **WriteList (Gemini)** — printer now uses `_sb.WriteList(extends, PrintType)`.
- **XML doc (Copilot)** — reworded to match the actual `IsTranspilable: false` predicate.

752/752 tests pass.

danfma:
Test ajustado para usar IList<T> em vez de IEnumerable<T> em 0d8d6c0. IEnumerable<T> deveria virar Iterable<T> (lazy sequence) e não Array<T>; o mapping atual está errado em IsCollectionLike. Abri #129 para corrigir o mapping em escopo separado.

PR #130, opened 2026-04-25
fix(core): map IEnumerable<T> to Iterable<T> instead of Array

### LINKED ISSUE #129: Fix IEnumerable<T> mapping to TypeScript Iterable<T> instead of Array<T>
## Problem

`System.Collections.Generic.IEnumerable<T>` currently maps to TypeScript `T[]` (via `IrArrayTypeRef`) because `RoslynTypeQueries.IsCollectionLike` (line 47-59) lumps it together with `IList<T>`, `ICollection<T>`, etc. The mapper at `IrTypeRefMapper.cs:176` then lowers it through `IrArrayTypeRef`.

## Why this is wrong

C# `IEnumerable<T>` is a lazy, read-only, forward-only sequence. The closest TypeScript surface is `Iterable<T>` (which `for...of` consumes), **not** `Array<T>`:
- `Array<T>` adds random access (`.length`, `[i]`), mutation (`.push`, `.pop`), and the full prototype.
- `Iterable<T>` is the minimal surface — exactly what `IEnumerable<T>` represents.

The codebase already correctly maps `IReadOnlyCollection<T>` → `Iterable<T>` at `IrTypeRefMapper.cs:168-173`. `IEnumerable<T>` should follow the same path.

## Symptom

```csharp
public interface ITyped<T>
{
    IEnumerable<T> Items { get; }
}
```

Today emits:
```ts
export interface ITyped<T> {
  readonly items: T[];     // wrong — implies random access + mutation
}
```

Should emit:
```ts
export interface ITyped<T> {
  readonly items: Iterable<T>;
}
```

The bug surfaces in property/parameter/return types and was made visible by the #117 review (PR #128) where using `IEnumerable<T>` as a base interface produced `extends T[]` (illegal heritage shorthand) — that PR worked around it by normalizing array shorthand to `Array<T>`, but the underlying mapping is still incorrect.

## Proposal

Move `IEnumerable<T>` out of `IsCollectionLike` and add a new branch to `IrTypeRefMapper` that maps it to `IrIterableTypeRef`, mirroring the existing `IReadOnlyCollection<T>` rule.

## Acceptance

- [ ] `IEnumerable<T>` parameters / returns / properties / fields emit as `Iterable<T>` (not `T[]`).
- [ ] LINQ-style usage paths (`.Where`, `.Select`, etc. on `IEnumerable<T>`) still lower correctly — the lazy LINQ runtime accepts `Iterable<T>` inputs.
- [ ] Sample regeneration (SampleTodo, SampleIssueTracker, SampleTodo.Service, SampleCounter) shows no semantic regressions; only `T[]` → `Iterable<T>` substitutions where appropriate.
- [ ] Existing golden tests reflect the new mapping; goldens that genuinely needed `Array` semantics (random access, mutation) switch to `IList<T>` / `IReadOnlyList<T>` on the C# side.

## Risks

- Many goldens will shift — the runtime still has to accept `Iterable<T>` everywhere `T[]` was tolerated. Verify via the bun test suites in `targets/js/sample-*` after regenerating.
- LINQ runtime entry points (`from`, `where`, etc.) need to keep accepting both `Iterable<T>` and arrays (arrays are iterables, so this should be free).

danfma:
## Summary

C# `IEnumerable<T>` is a lazy, read-only, forward-only sequence. The closest TypeScript surface is `Iterable<T>` — what `for...of` consumes — **not** `Array<T>` (which adds random access, mutation, and the full prototype). The mapping was incorrect because `RoslynTypeQueries.IsCollectionLike` lumped `IEnumerable` together with `IList`, `ICollection`, etc., and the IR mapper lowered all of them through `IrArrayTypeRef` (TS `T[]`).

## Mechanism

- Drop `System.Collections.Generic.IEnumerable` from `IsCollectionLike`.
- Add a dedicated branch in `IrTypeRefMapper.MapNamed` that handles both `IEnumerable<T>` and the existing `IReadOnlyCollection<T>` mapping → `IrIterableTypeRef` (printer emits `Iterable<T>`).
- Iterator (yield-bearing) methods unchanged: they route through `MapForGeneratorReturn` and emit `Generator<T>`.

## Symptom (now fixed)

```csharp
public IEnumerable<int> Snapshot() => [1, 2, 3];
```

Before: `snapshot(): number[]` (wrong — implies random access + mutation)
After:  `snapshot(): Iterable<number>` (matches the C# semantics)

The bug was visible in PR #128 reviews, where using `IEnumerable<T>` as a base interface produced `extends T[]` (illegal heritage shorthand). #128 worked around the symptom by rewriting `T[]` to `Array<T>` in heritage clauses; this PR fixes the root cause for all positions (parameters, return types, properties, fields, base interfaces).

## Test plan

- [x] Updated existing tests that asserted the old (wrong) behavior:
  - `YieldTranspileTests.IEnumerableWithoutYield_RemainsArrayMapping` → `IEnumerableWithoutYield_MapsToIterable`.
  - `ExtensionTranspileTests.ClassicExtension_GenericReceiver` — expects `Iterable<T>` instead of `T[]`.
- [x] Full .NET suite: 752/752 pass.
- [x] Sample regeneration produced byte-identical output (no sample uses `IEnumerable<T>` directly today, only `IList`/`IReadOnlyList`).
- [x] Bun test suites:
  - `metano-runtime`: 275 pass
  - `sample-todo`: 18 pass
  - `sample-issue-tracker`: 142 pass
  - `sample-todo-service`: 33 pass

Closes #129

### DIFF
--- a/src/Metano.Compiler.TypeScript/Transformation/JsonSerializerContextTransformer.cs
+++ b/src/Metano.Compiler.TypeScript/Transformation/JsonSerializerContextTransformer.cs
@@ -426,8 +426,22 @@ public sealed class JsonSerializerContextTransformer(TypeScriptTransformContext 
             if (named.IsSetLike() && named.TypeArguments.Length > 0)
                 return "hashSet";
 
-            // Collection-like → array
-            if (named.IsCollectionLike() && named.TypeArguments.Length > 0)
+            // Collection-like / IEnumerable / IReadOnlyCollection → array.
+            // Since #129 these no longer share the `IsCollectionLike`
+            // predicate (they lower to TS `Iterable<T>` at the type
+            // level), but JSON-wise they still serialize as arrays —
+            // `JSON.stringify([...iter])` works the same as a plain
+            // array. Keep the descriptor as "array" so the generated
+            // serializer context knows to iterate and emit the element
+            // descriptor.
+            if (
+                (
+                    named.IsCollectionLike()
+                    || fullName.StartsWith("System.Collections.Generic.IEnumerable")
+                    || fullName.StartsWith("System.Collections.Generic.IReadOnlyCollection")
+                )
+                && named.TypeArguments.Length > 0
+            )
                 return "array";
 
             // Enum with [StringEnum] → enum, otherwise numericEnum
--- a/src/Metano.Compiler/Extraction/IrTypeRefMapper.cs
+++ b/src/Metano.Compiler/Extraction/IrTypeRefMapper.cs
@@ -165,9 +165,17 @@ public static class IrTypeRefMapper
                     Map(named.TypeArguments[1], originResolver)
                 );
 
-            // IReadOnlyCollection<T> → Iterable
+            // IEnumerable<T> / IReadOnlyCollection<T> → Iterable.
+            // C# `IEnumerable<T>` is a lazy, read-only, forward-only
+            // sequence — the closest TS surface is `Iterable<T>`,
+            // not `Array<T>`. Iterator return types (yield) take a
+            // separate path through `MapForGeneratorReturn` and
+            // lower to `Generator<T>` instead.
             if (
-                fullName.StartsWith("System.Collections.Generic.IReadOnlyCollection")
+                (
+                    fullName.StartsWith("System.Collections.Generic.IEnumerable")
+                    || fullName.StartsWith("System.Collections.Generic.IReadOnlyCollection")
+                )
                 && named.TypeArguments.Length > 0
             )
                 return new IrIterableTypeRef(Map(named.TypeArguments[0], originResolver));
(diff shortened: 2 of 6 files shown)


danfma:
@chatgpt-codex-connector @gemini-code-assist thanks for the review. Both findings addressed in 6e6bb15.

- **Codex P1 — JSON classification regression.** Genuine bug. `JsonSerializerContextTransformer.ClassifyPropertyType` previously routed `IEnumerable<T>` properties through `IsCollectionLike` and emitted `kind: "array"`. After dropping `IEnumerable` from the predicate, the classifier silently fell through to `kind: "ref"`, breaking generated serializer contexts for `[JsonSerializable]` types with iterable properties. Added an explicit branch covering both `IEnumerable<T>` and `IReadOnlyCollection<T>` so JSON-wise they keep array semantics. Locked in by `IEnumerableProperty_ClassifiesAsArrayDescriptor`.
- **Gemini — `target` propagation.** Threaded the `target` parameter through the new iterable branch's recursive `Map` call so `[Name(target, ...)]` overrides on the type argument are honored. Same fix applied to the array branch one line below.

753/753 tests pass.

Issue #133, opened 2026-04-25
Method group .bind(receiver) re-evaluates non-pure receivers


danfma:
## Problem

When the IR lowers a method group → delegate assignment, it now emits \`receiver.method.bind(receiver)\` (PR fixing #132). For receivers with side effects — e.g. \`Cache().Service.Handler\` — the receiver subexpression is **emitted twice** and the body executes twice.

\`\`\`csharp
target.OnTick = Cache().Service.Tick;
\`\`\`

emits:

\`\`\`ts
target.onTick = cache().service.tick.bind(cache().service);
//                                       ^^^^^^^^^^^^^^^^^ second evaluation
\`\`\`

The first \`cache()\` resolves the function reference; the bind argument re-runs \`cache()\` and walks \`.service\` again.

## Why this is OK to defer

Pure receivers dominate real use: \`this\`, identifiers, plain field/property chains. The double-eval was already present on the pre-existing \`[This]\` path; PR #132 expands the surface area but doesn't introduce a new class of bug.

## Proposal

When \`IrMemberAccess.Target\` is not a pure shape (identifier, \`this\`, \`base\`, \`IrTypeReference\`, or chain of those), wrap the bind site in an immediately-invoked arrow that captures the receiver in a temporary:

\`\`\`ts
target.onTick = ((r) => r.tick.bind(r))(cache().service);
\`\`\`

or introduce an \`IrLet\`-style IR shape that the printer lowers into a temp variable. The let-binding form is more general and worth the IR change if other lowerings can reuse it.

## Acceptance

- [ ] Method group bind on a side-effecting receiver chain evaluates the chain exactly once.
- [ ] Existing pure-receiver paths still emit the simple \`obj.method.bind(obj)\` shape (no IIFE overhead when not needed).
- [ ] Test pinning \`Cache().Service.Tick\` shape and counting receiver evaluations.

## Related

- #132 — surfaced this when the wrapper started covering all delegate targets.

PR #134, opened 2026-04-25
fix(ts): bind 'this' on method group conversion to delegate

### LINKED ISSUE #132: Method group assignment to delegate field loses 'this' binding in TypeScript
## Problem

When a C# method group is assigned to a delegate-typed field/property, the C# compiler implicitly captures \`this\` so the delegate can be invoked from any call site and still resolve member accesses on the original receiver. The current TypeScript emission preserves the method-group assignment as a bare property reference, which produces a JS function whose \`this\` binding is lost at the call site — invoking the delegate later resolves \`this\` to whatever object holds the field.

## Reproduction

C# input:

\`\`\`csharp
using SampleCounterWithDomView.Models;
using SampleCounterWithDomView.Views;

namespace SampleCounterWithDomView.Presenters;

public sealed class CounterPresenter(ICounterView view, Counter initialState)
{
    private readonly ICounterView _view = view;
    private Counter _state = initialState;

    public void StartApplication(string containerName)
    {
        var renderer = new Renderer(containerName);
        renderer.Render(_view);

        _view.OnButtonClick = OnButtonClicked;
        _view.ShowCounter(_state.Count);
    }

    private void OnButtonClicked()
    {
        _state = _state.Increment();
        _view.ShowCounter(_state.Count);
    }
}
\`\`\`

Current TS output:

\`\`\`ts
startApplication(containerName: string): void {
  const renderer = new Renderer(containerName);
  renderer.render(this._view);
  this._view.onButtonClick = this.onButtonClicked;     // bug: \`this\` not bound
  this._view.showCounter(this._state.count);
}
\`\`\`

When \`_view.onButtonClick()\` later fires, JS resolves \`this\` to \`_view\`, so \`this._state.Increment()\` inside \`onButtonClicked\` throws (\`_state\` does not exist on the view).

## Expected

The transpiler must emit a bound reference when a method group is assigned to a delegate field/property. Either form is correct; pick whichever composes best with the existing delegate runtime helpers (\`delegateAdd\` / \`delegateRemove\`):

\`\`\`ts
this._view.onButtonClick = this.onButtonClicked.bind(this);
// or
this._view.onButtonClick = () => this.onButtonClicked();
\`\`\`

The same fix applies in every position where a method group flows into a delegate-typed slot: assignment, \`return method;\`, argument passing, object-literal property, event \`+=\` (already routes through \`delegateAdd\` — verify no double-binding), constructor field initializer.

## Detection rule

Method group conversion triggers when:
1. The expression is a \`MemberAccessExpression\` (or unqualified identifier) whose symbol is an \`IMethodSymbol\`,
2. The expected type at that position is a delegate type (\`INamedTypeSymbol.TypeKind == TypeKind.Delegate\`),
3. The receiver is an instance (static methods do not need binding).

## Acceptance

- [ ] Method group → delegate emits \`.bind(this)\` (or equivalent arrow wrap) preserving the original receiver.
- [ ] Static method groups skip the bind (no \`this\` to capture).
- [ ] Event subscriptions (\`evt += handler\`) keep working — \`delegateAdd\` should receive an already-bound handler.
- [ ] Method-group passing to a delegate parameter (e.g., \`Run(OnTick)\`) is also bound.
- [ ] Tests:
  - Assignment to a delegate field on another object (the case above).
  - Method group passed as a delegate argument to a method.
  - Static method group (no bind).
  - \`+=\` event handler still resolves \`this\` correctly inside the handler body.
  - Roundtrip Bun test: assign \`OnButtonClick = OnButtonClicked\`, invoke from the view, verify \`this\` inside the handler points at the presenter.

## Related

Surfaced by the SampleCounter / SampleCounterWithDomView consumer.

danfma:
## Summary

C# method groups assigned to a delegate slot capture the instance receiver at the conversion site so any later invocation runs the body with the original \`this\`. JS doesn't — a bare \`obj.method\` reference loses its receiver as soon as it leaves the access expression. The extractor's wrapper previously only fired for delegates whose first parameter carried \`[This]\`, so ordinary delegates (\`Action\`, \`Func\`, custom delegates without \`[This]\`) emitted the bare reference and broke \`this\` inside the body whenever the delegate was invoked through a different object.

Repro from the SampleCounterWithDomView consumer:

\`\`\`csharp
_view.OnButtonClick = OnButtonClicked;
\`\`\`

Before: \`this._view.onButtonClick = this.onButtonClicked;\` — \`this\` inside \`OnButtonClicked\` resolves to \`_view\` at call time, breaking \`this._state.Increment()\`.

After: \`this._view.onButtonClick = this.onButtonClicked.bind(this);\`

## Mechanism

- Rename \`WrapMethodGroupForThisDelegate\` → \`WrapMethodGroupForDelegate\`.
- Split the wrap into two independent steps:
  - Always emit \`.bind(receiver)\` for instance method groups assigned to any delegate slot. Static methods, type-qualified references, and bare-identifier shapes (locals, local functions) skip the bind — no instance to capture.
  - Stack \`bindReceiver(...)\` on top only when the delegate's first parameter carries \`[This]\`. That contract is unchanged; the trampoline still rebinds JS \`this\` into the method's first parameter.
- Dart target opts out via a new \`IsDartTarget\` guard. Dart tear-offs auto-bind, so the JS-only helpers would only get in the way.
- Extracted \`ShouldBindInstanceReceiver\` predicate for readability (Bob's clean-code review).

## Test plan

8 new tests in \`DelegateEventTests.cs\` cover:

- Assignment, return statement, argument passing — all bind \`.bind(this)\`.
- Static method group — skips bind.
- Ordinary call site — stays un-bound (locks the \`ConvertedType\` invariant).
- \`[This]\` + instance method group — stacks \`bindReceiver(...bind(this))\`.
- Dart opt-out — neither \`.bind\` nor \`bindReceiver\` appears.
- Event \`+= handler\` — \`event\$add(handler.bind(this))\`, no double-binding.

\`.DoesNotContain("bindReceiver")\` assertions on every non-\`[This]\` case pin the no-trampoline invariant.

- [x] Full .NET suite: 761/761 pass.
- [x] Sample regeneration: byte-identical (no sample uses method-group delegate assignment today).
- [x] Bun suites: 275 (runtime) + 18 (todo) + 142 (issue-tracker) + 33 (todo-service) green.
- [x] csharpier-format clean.

## Known limitation

\`.bind(receiver)\` evaluates the receiver subexpression twice when the receiver chain has side effects (\`Cache().Service.Tick\`). Pre-existing concern under the \`[This]\` path; the expanded surface area is tracked separately in #133.

- compiler-man: 4 Major test gaps (Dart opt-out, ordinary call site, \`[This]\`+instance stacking, no-double-bind assertions) — all locked in.
- bob: Clean-code polish — extracted \`IsDartTarget\` property and \`ShouldBindInstanceReceiver\` predicate, added blank-line spacing between early returns, switched the trailing if/return into a single ternary.

Closes #132
Part of #133

### DIFF
--- a/src/Metano.Compiler/Extraction/IrExpressionExtractor.cs
+++ b/src/Metano.Compiler/Extraction/IrExpressionExtractor.cs
@@ -581,7 +581,7 @@ public sealed class IrExpressionExtractor
             && symbol.ContainingType is not null
             && !IsLocalLikeSymbol(symbol)
         )
-            return WrapMethodGroupForThisDelegate(
+            return WrapMethodGroupForDelegate(
                 generic,
                 symbol,
                 new IrMemberAccess(
@@ -595,7 +595,7 @@ public sealed class IrExpressionExtractor
             symbol is { IsStatic: true } and (IPropertySymbol or IFieldSymbol or IMethodSymbol)
             && symbol.ContainingType is not null
         )
-            return WrapMethodGroupForThisDelegate(
+            return WrapMethodGroupForDelegate(
                 generic,
                 symbol,
                 new IrMemberAccess(
@@ -605,7 +605,7 @@ public sealed class IrExpressionExtractor
                 )
             );
 
-        return WrapMethodGroupForThisDelegate(
+        return WrapMethodGroupForDelegate(
             generic,
             symbol,
             new IrIdentifier(generic.Identifier.ValueText)
@@ -642,7 +642,7 @@ public sealed class IrExpressionExtractor
                 id.Identifier.ValueText,
                 BuildOrigin(symbol)
             );
-            return WrapMethodGroupForThisDelegate(id, symbol, memberAccess);
+            return WrapMethodGroupForDelegate(id, symbol, memberAccess);
         }
 
         // Static member of the enclosing (or any other) type reached without a
@@ -660,14 +660,10 @@ public sealed class IrExpressionExtractor
                 id.Identifier.ValueText,
                 BuildOrigin(symbol)
             );
-            return WrapMethodGroupForThisDelegate(id, symbol, staticAccess);
+            return WrapMethodGroupForDelegate(id, symbol, staticAccess);
         }
 
-        return WrapMethodGroupForThisDelegate(
-            id,
-            symbol,
-            new IrIdentifier(id.Identifier.ValueText)
-        );
+        return WrapMethodGroupForDelegate(id, symbol, new IrIdentifier(id.Identifier.ValueText));
     }
 
     private static bool IsLocalLikeSymbol(ISymbol symbol) =>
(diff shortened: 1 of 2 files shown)


danfma:
@gemini-code-assist @copilot-pull-request-reviewer @chatgpt-codex-connector thanks. All four findings addressed in c4a4e9b.

- **`base.Method` produces invalid `super.method.bind(super)` (Gemini HIGH, Copilot, Codex P1).** Genuine bug introduced by broadening the bind path. Fixed by substituting `IrThisExpression` for the bind argument when the receiver is `IrBaseExpression`. The base method's body still runs against the current instance — matches C#'s `base.Method` semantics (non-virtual dispatch but receiver stays `this`). Encapsulated as `BindArgumentFor(IrExpression)` so the rule has a single documented home. Locked in by `BaseMethodGroup_BindsThisInsteadOfSuper`.
- **Inline `[This]` check duplication (Gemini medium).** Collapsed into the existing `HasThisParameter(IMethodSymbol)` helper for consistency with the two other call sites in the same file.

The receiver double-evaluation note is acknowledged in the PR description and tracked separately as #133.

762/762 tests pass.

PR #135, opened 2026-04-27
chore(release): route conventional commits to changelog categories

danfma:
## Summary

dotnet-releaser's built-in auto-labelers expect imperative-English commit titles (\"Add foo\", \"Fix bar\"), but this repo uses Conventional Commits (\`feat:\`, \`fix:\`, \`chore:\`, etc.). Without explicit auto-labelers, every non-\`Fix\`-prefixed title falls through to the catch-all \`misc\` label and every feature lands under \"🧰 Misc\" — see [v0.8.0 release notes](https://github.com/danfma/metano/releases/tag/v0.8.0).

## Mechanism

Custom \`[[changelog.autolabeler]]\` entries run **before** the built-in defaults. Each pattern accepts the Conventional Commits shape \`<type>(<scope>)?(!)?:\` and routes the commit to the corresponding built-in category:

| Prefix | Section |
|---|---|
| \`feat:\` | ✨ New Features |
| \`fix:\` | 🐛 Bug Fixes |
| \`refactor:\` / \`perf:\` / \`style:\` | 🚀 Enhancements |
| \`chore:\` / \`build:\` / \`ci:\` | 🧰 Maintenance |
| \`test:\` | 🏭 Tests |
| \`docs:\` | 📚 Documentation |
| \`<type>!:\` (any type with breaking marker) | 🚨 Breaking Changes |

The \`misc\` catch-all (\`title = '.'\`) stays in place via the defaults, absorbing anything that doesn't match a Conventional-Commits prefix.

## Verification

Config-only change. Validation will happen on the next release tag — the v0.9.0 changelog should split feature/maintenance commits across the right sections instead of bundling them under Misc. If a commit type is missed, the fix is one extra autolabeler entry.

### DIFF
--- a/dotnet-releaser.toml
+++ b/dotnet-releaser.toml
@@ -26,3 +26,43 @@ enable = false
 # Coverage is not yet supported for Testing Platform (TUnit) projects.
 [coverage]
 enable = false
+
+# ── Changelog ────────────────────────────────────────────────────
+#
+# We use Conventional Commits (`feat:`, `fix:`, `chore:`, etc.) as the
+# canonical commit format. dotnet-releaser's built-in auto-labelers
+# expect imperative-English titles ("Add foo", "Fix bar"), so without
+# the entries below `feat:` and friends would fall through to the
+# catch-all `misc` label and every feature would land under "🧰 Misc"
+# in the release notes (which is what happened up to v0.8.0).
+#
+# Custom auto-labelers run BEFORE defaults, so listing them here
+# routes our commit titles to the appropriate categories. Each
+# pattern accepts an optional scope (`feat(ts):`, `fix(core):`).
+[[changelog.autolabeler]]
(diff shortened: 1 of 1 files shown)


danfma:
@gemini-code-assist thanks. Both points addressed in c922c0a:

- **Greedy scope (`(.+)`)**: replaced with `([^)]+)` on every autolabeler so a stray colon or parenthesis later in the title cannot push the regex past the closing paren of the scope.
- **Missing breaking-change marker (`!`)**: every type-prefixed autolabeler now accepts an optional `!?` before the colon. Combined with the existing dedicated `breaking-change` autolabeler, a breaking feature (`feat!:` / `feat(ts)!:`) lands in both the ✨ New Features and 🚨 Breaking Changes sections — dotnet-releaser fires every matching autolabeler, so multiple labels per commit are supported.

PR #138, opened 2026-04-27
fix(ts): make package.json writes additive, not destructive

### LINKED ISSUE #136: package.json fields overwritten on every build instead of write-if-missing
## Problem

Every transpiler run unconditionally rewrites several \`package.json\` fields, even when the consumer has already set them by hand. The expected behavior — and what the docstring on \`PackageJsonWriter.UpdateOrCreate\` claims (\"non-destructive merge\") — is that the transpiler should only write fields that are missing from the existing \`package.json\`.

Today's behavior diverges in concrete ways:

- \`src/Metano.Compiler.TypeScript/PackageJsonWriter.cs:167\` — \`root[\"type\"] = \"module\"\` runs unconditionally. A consumer who deliberately set \`\"type\": \"commonjs\"\` (e.g. for a Node CJS dual-build sample) gets it flipped back to \`\"module\"\` on every regeneration.
- \`PackageJsonWriter.cs:168\` — \`root[\"sideEffects\"] = false\` runs unconditionally. Consumers who curate this field (e.g. listing CSS imports as side-effectful for tree-shakers) lose their value.
- \`PackageJsonWriter.cs:178-194\` — the \`exports\` object is *fully* replaced when the transpiler emits any exports. The current comment claims this is intentional (\"transpiler-controlled\"), but per the new convention the transpiler should only seed missing keys; user-added subpaths should survive.
- \`PackageJsonWriter.cs:163\` — \`root[\"name\"]\` is rewritten when \`[assembly: EmitPackage(...)]\` declares a name. There is already a divergence warning, but the rule per this issue is that the transpiler should not overwrite an existing \`name\` at all (\`EmitPackage\` should only seed it on first creation, then warn if they later diverge).

Imports (\`#\` and \`#/*\`) and \`dependencies\` already merge correctly — those code paths can stay as-is.

## Reproduction

1. Generate any sample project: \`dotnet run --project src/Metano.Compiler.TypeScript/ -- -p samples/SampleTodo/SampleTodo.csproj -o targets/js/sample-todo/src --clean\`.
2. Open \`targets/js/sample-todo/package.json\`, change \`\"type\": \"module\"\` → \`\"type\": \"commonjs\"\` (or set \`\"sideEffects\": [\"./styles.css\"]\`).
3. Re-run the transpiler.
4. Observe: \`type\` is back to \`module\`, \`sideEffects\` is back to \`false\`. The hand-edit was lost without a diagnostic.

## Proposed rule

For *every* package.json field the transpiler touches, follow a single policy:

- **If the field is missing in the existing file** → write the transpiler's value as the seed.
- **If the field is present** → leave it alone, even if it differs from what the transpiler would emit.
- **Optional**: emit an \`MS00xx\` informational diagnostic when an existing value diverges from what the transpiler would have emitted, so the consumer knows the customization is in effect.

Exceptions where the transpiler genuinely owns the content:

- \`imports\` keys \`#\` and \`#/*\` (managed aliases — already replace-on-emit, keep that behavior).
- \`dependencies\` entries that come from \`[EmitPackage]\` cross-package resolution (additive merge — already correct).

## Acceptance

- [ ] \`type\`, \`sideEffects\`, and \`name\` written only when missing.
- [ ] \`exports\` keys merge additively — user-added subpaths survive across regenerations; transpiler keys still get refreshed.
- [ ] Existing tests for sample regeneration still pass with no diff under \`targets/js/*\` (samples don't customize these fields today, so the change should be silent for the happy path).
- [ ] New tests:
  - Existing \`type: \"commonjs\"\` survives a regeneration.
  - Existing \`sideEffects: [\"./styles.css\"]\` survives a regeneration.
  - User-added \`exports[\"./extras\"]\` entry survives.
  - Missing fields still get seeded on first generation.

## Where to look

- \`src/Metano.Compiler.TypeScript/PackageJsonWriter.cs\` — single function (\`UpdateOrCreate\`) holds all the writes.
- \`tests/Metano.Tests/PackageJsonWriterTests.cs\` (if present) or the tests under \`tests/Metano.Tests/\` that exercise sample regeneration.

## Related

Surfaced by user feedback after observing local edits to \`package.json\` getting reverted on rebuild.

danfma:
## Summary

Every transpiler run unconditionally rewrote \`type\`, \`sideEffects\`, \`name\`, and the \`exports\` object — destroying hand-curated values on every regeneration. The docstring already called the merge \"non-destructive\", but the implementation only honored that for \`imports\` and \`dependencies\`.

Switch all written fields to a write-if-missing rule:

- **\`type\` / \`sideEffects\`** — seeded with \`module\` / \`false\` only when absent. A consumer who set \`type: \"commonjs\"\` (CJS dual build) or \`sideEffects: [\"./styles.css\"]\` (tree-shaker hint) keeps that value across regenerations.
- **\`name\` (with \`[EmitPackage]\`)** — seeded only on first creation. Existing names are preserved; **MS0007 still fires on divergence** so the consumer notices the mismatch but published links and cross-package imports do not break.
- **\`exports\`** — additive merge with two layers:
  - **Stale removal**: an existing entry whose key is not in the generated set is dropped only when the value matches the transpiler-emitted shape (\`{types, import}\` strings both pointing into \`./{dist}/...\`). User-added subpaths (\`./styles.css\`, declaration shims) survive.
  - **Per-key deep-merge**: when a generated key is already present, only the transpiler's \`types\` / \`import\` fields refresh in place. Extra conditional fields the consumer added (e.g. \`require\` for CJS dual builds) stay untouched.

## Mechanism

\`UpdateOrCreate\` body now reads as a sequence of named steps. Policy decisions live in dedicated helpers:

- \`ApplyAuthoritativePackageName\` — name preservation + MS0007.
- \`WriteIfMissing\` — scalar seeding for \`type\`/\`sideEffects\`.
- \`MergeTranspilerManagedExports\` — additive merge entry point.
- \`DropStaleTranspilerEntries\` + \`MergeGeneratedEntries\` — the two layers of the exports merge.
- \`IsTranspilerEmittedExportEntry\` — shape predicate used by stale removal.

**compiler-man** flagged a Major: the prior version unconditionally overwrote generated keys via \`existing[key] = generated[key].DeepClone()\`, dropping user-augmented conditional fields. Fixed by deep-merging matching keys (only refresh transpiler-emitted fields; preserve extras).

**bob** flagged the body length and naming. Extracted \`ApplyAuthoritativePackageName\`, renamed the predicate to \`IsTranspilerEmittedExportEntry\`, and split the exports merge into two single-purpose helpers.

## Test plan

- [x] 5 new tests in \`EmitPackageTests.cs\`:
  - \`type: \"commonjs\"\` survives a regeneration; missing \`type\` is seeded as \`module\`.
  - \`sideEffects: [\"./styles.css\", ...]\` survives.
  - User-added \`./styles.css\` export survives.
  - User-augmented \`require\` field on a transpiler key survives while \`types\`/\`import\` refresh.
  - Executable consumer's existing \`exports\` object is preserved (was unconditionally removed before).
- [x] Renamed \`ExistingFileWithDivergentName_WarnsAndOverwrites\` → \`...WarnsAndPreservesExisting\` to reflect the new contract.
- [x] Existing \`Exports_StaleEntriesRemovedOnRegeneration\` still passes — the shape heuristic correctly classifies \`./old-barrel\` (transpiler-shaped) and drops it.
- [x] Full .NET suite: 768/768 pass.
- [x] csharpier-format clean.

Closes #136

### DIFF
--- a/src/Metano.Compiler.TypeScript/PackageJsonWriter.cs
+++ b/src/Metano.Compiler.TypeScript/PackageJsonWriter.cs
@@ -143,54 +143,24 @@ public static class PackageJsonWriter
             };
         }
 
-        // [EmitPackage] is the source of truth for the package name when present.
-        if (authoritativePackageName is not null)
-        {
-            var existingName = root["name"]?.GetValue<string>();
-            if (existingName is not null && existingName != authoritativePackageName)
-            {
-                diagnostics.Add(
-                    new MetanoDiagnostic(
-                        MetanoDiagnosticSeverity.Warning,
-                        DiagnosticCodes.CrossPackageResolution,
-                        $"package.json#name '{existingName}' diverges from "
-                            + $"[assembly: EmitPackage(\"{authoritativePackageName}\")]. "
-                            + $"Overwriting with the attribute value — consumers will import via "
-                            + $"'{authoritativePackageName}'."
-                    )
-                );
-            }
-            root["name"] = authoritativePackageName;
-        }
+        ApplyAuthoritativePackageName(root, authoritativePackageName, diagnostics);
 
-        // Apply controlled fields
-        root["type"] = "module";
-        root["sideEffects"] = false;
+        // Seed defaults when the consumer has not picked their own
+        // values. The transpiler must not flip a hand-curated `type`
+        // or `sideEffects` value back on every regeneration.
+        WriteIfMissing(root, "type", "module");
+        WriteIfMissing(root, "sideEffects", false);
 
-        // Replace transpiler-managed entries while preserving user-defined ones.
-        // For imports, the transpiler manages "#" and "#/*" — stale aliases are removed.
+        // Imports: the transpiler owns `#` and `#/*` — stale aliases
+        // are removed; user-defined keys survive.
         HashSet<string> managedImportKeys = ["#", "#/*"];
         ReplacePreservingUserEntries(root, "imports", imports, managedImportKeys);
 
-        // Exports are fully transpiler-controlled — replace entirely.
-        if (exports is not null)
+        // Exports merge additively — see `MergeTranspilerManagedExports`
+        // for the shape-detection rule and the deep-merge contract.
+        if (exports is { Count: > 0 })
         {
-            if (root["exports"] is JsonObject existingExports)
-            {
-                // Replace in-place to preserve key ordering.
-                foreach (var k in existingExports.Select(e => e.Key).ToList())
-                    existingExports.Remove(k);
-                foreach (var (ek, ev) in exports)
-                    existingExports[ek] = ev?.DeepClone();
-            }
-            else
-            {
-                root["exports"] = exports;
-            }
-        }
-        else
-        {
-            root.Remove("exports");
+            MergeTranspilerManagedExports(root, exports, distDirRelativeToPackageRoot);
         }
 
         // Merge auto-generated cross-package dependencies into the existing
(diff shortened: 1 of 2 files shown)


danfma:
@gemini-code-assist @copilot-pull-request-reviewer @chatgpt-codex-connector thanks. All three findings addressed in 3d9fc9a — three reviewers independently flagged two of them.

- **Path-boundary on dist prefix (Gemini HIGH, Copilot, Codex P2)**: Bare `./dist` matched `./dist-cjs/...` etc. via `StartsWith`. Appended trailing `/` so the comparison anchors at the directory boundary. Locked in by `SiblingDistDirectoryExport_NotMisclassifiedAsStale`.
- **Empty-exports stale removal (Gemini medium, Copilot, Codex P1)**: The `Count > 0` guard skipped the merge when the current generation produced no barrels, leaving stale `.`/`./domain` from a previous run. Switched to `is not null` so the merge runs for non-executable builds; user-added subpaths still survive thanks to the shape filter. Locked in by `LibraryWithoutBarrels_PrunesStaleTranspilerEntries`.
- **Defensive string read (Gemini medium)**: `JsonValue.GetValue<string>()` throws on non-scalar nodes. Hand-edited `package.json` may legally hold any JSON shape under `types`/`import`. Extracted `ReadStringField` returning `null` for non-string nodes so the writer classifies the entry as user-curated and preserves it. Locked in by `NonStringExportFields_DoNotCrashWriter`.

771/771 tests pass.

PR #139, opened 2026-04-27
fix(runtime): declare metano-runtime tree-shakeable

### LINKED ISSUE #137: Runtime not tree-shakeable — 175KB bundle for trivial sample
## Problem

\`sample-counter-with-dom-view\` has no external/DOM dependencies and only imports from \`metano-runtime\`. The Vite production build still ships **175KB** of JS at \`./targets/js/sample-counter-with-dom-view/dist/assets/index-C31QbsRF.js\` — orders of magnitude more than what the consumer actually uses.

A correctly tree-shakeable runtime should leave only the helpers the sample touches in the bundle. The 175KB figure suggests the bundler is pulling in the entire \`metano-runtime\` (Temporal polyfill, Decimal, HashSet, LINQ, JSON converters, etc.) even though the sample uses none of them.

## Root cause hypotheses

Each one is plausible on its own; likely a combination is in play.

### 1. Missing \`sideEffects\` flag in \`metano-runtime/package.json\`

Verified: the runtime's \`package.json\` does **not** declare \`\"sideEffects\": false\`. Without it, Webpack / Rollup / Rspack / esbuild conservatively assume every module in the package may have side effects and skip the tree-shake pass entirely. This alone is likely responsible for the bulk of the 175KB.

The transpiler-managed code path in \`PackageJsonWriter.cs:168\` *does* force \`sideEffects: false\` on **consumer** packages, but the runtime itself is hand-authored and missing the flag.

### 2. Top-level barrel re-exports without \`/*#__PURE__*/\` annotations

\`targets/js/metano-runtime/src/index.ts\` is two lines:

\`\`\`ts
export * from \"./system\";
export * from \"./type-checking\";
\`\`\`

Wildcard re-exports are themselves tree-shakeable when (a) every leaf module is side-effect-free and (b) the package's \`sideEffects\` flag tells the bundler so. Without (b), the entire barrel collapses into the bundle. Even with (b), the chain \`index.ts\` → \`system/index.ts\` → \`system/<helpers>\` must stay clean — any module that runs code at top level (registering polyfills, mutating prototypes, side-effectful module init) will pin the whole chain.

### 3. \`@js-temporal/polyfill\` on the dep path

\`metano-runtime\` declares \`@js-temporal/polyfill\` as a runtime dependency and imports it from \`system/temporal-helpers.ts\`. Polyfills typically execute side effects at top level (registering globals). If the bundler can't prove the import is unused, the entire polyfill (~70KB minified+gzipped) ends up in the bundle. The sample does not use Temporal, so this should be fully removable.

### 4. JSON converters drag in Decimal.js

\`system/json/types.ts\`, \`system/json/converters.ts\`, and \`system/json/index.ts\` reference Decimal. If \`metano-runtime/index.ts\` re-exports JSON converters via \`export * from \"./system\"\` and the JSON entry chains into Decimal, every consumer pulls Decimal even when only a counter helper is needed.

### 5. tsgo emit shape

The runtime is built with \`tsgo -b .\`. Worth checking the emitted \`.js\` files: if any of them have CommonJS-flavored output, top-level \`Object.defineProperty\` calls, or similar non-static patterns, bundlers can't statically analyze them and drop dead branches. Pure ESM with named exports is required for full tree-shaking.

### 6. Consumer-side bundler config

Vite + Rollup tree-shake aggressively by default, but \`build.lib\` mode, custom \`rollupOptions.preserveModules\`, or \`externalLiveBindings: true\` can suppress it. Worth dumping \`vite.config.ts\` for the consumer.

## Investigation steps (in order, by ROI)

1. **Add \`\"sideEffects\": false\`** to \`metano-runtime/package.json\`. Re-bundle. Measure the delta. If the bundle shrinks dramatically, this was the dominant issue.
2. **Run \`vite build --mode production --debug\`** in \`sample-counter-with-dom-view\` and inspect the bundler's \"used vs unused exports\" output (Rollup logs or \`rollup-plugin-visualizer\`). Identifies which modules are landing in the bundle.
3. **Move \`@js-temporal/polyfill\` import behind a lazy boundary** if it stays in the bundle even after sideEffects flag — e.g., dynamic \`import(\"@js-temporal/polyfill\")\` inside the helpers that use it, so static analysis can prove it's unused.
4. **Audit \`system/index.ts\` and \`type-checking/index.ts\`** for top-level statements (constants that aren't \`const\`, IIFEs, side-effectful registers).
5. **Sub-package the heavyweight modules** if needed: \`metano-runtime/json\` and \`metano-runtime/temporal\` as separate entry points so the main barrel doesn't transitively pull them. \`exports\` already supports \`./system/json\` as a sub-entry.

## Acceptance

- [ ] Bundle size for \`sample-counter-with-dom-view\` drops to \`< 20KB\` (counter helpers + minimal runtime). Target is \"essentially zero runtime overhead beyond what the sample actually calls\".
- [ ] Bundle visualizer shows no Temporal polyfill, no Decimal, no JSON converters, no LINQ, no HashSet code in the final output.
- [ ] Other samples (\`sample-todo\`, \`sample-issue-tracker\`, \`sample-todo-service\`) keep building and their bundle sizes do not regress (they use more of the runtime, but should still drop unused parts).

## Where to look

- \`targets/js/metano-runtime/package.json\` — missing \`sideEffects\`.
- \`targets/js/metano-runtime/src/index.ts\` — barrel re-exports.
- \`targets/js/metano-runtime/src/system/index.ts\` (and its barrel chain).
- \`targets/js/metano-runtime/src/system/temporal-helpers.ts\` — Temporal polyfill import.
- \`targets/js/metano-runtime/src/system/json/{converters,types,index}.ts\` — Decimal use.
- \`targets/js/sample-counter-with-dom-view/vite.config.ts\` (if any) — bundler config.
- \`src/Metano.Compiler.TypeScript/PackageJsonWriter.cs:168\` — already sets consumer-side \`sideEffects: false\`. Confirms the convention; runtime should follow suit.

## Related

Discovered while measuring \`sample-counter-with-dom-view\` build output. Same investigation will inform whether \`PackageJsonWriter\` should also auto-set \`sideEffects: false\` on the runtime's own package.json (likely no — runtime is hand-authored and not regenerated).

danfma:
## Summary

\`metano-runtime/package.json\` was missing \`\"sideEffects\": false\`. Bundlers conservatively assumed each module might run side-effectful code at top level and disabled tree-shaking for the whole package, pulling the entire runtime (Temporal polyfill, Decimal, HashSet, LINQ, JSON converters) into the consumer's bundle even when the consumer only imported \`HashCode\`.

## Verification

\`bun build\` against a minimal entry that imports only \`HashCode\`:

\`\`\`js
import { HashCode } from \"metano-runtime\";

const hc = new HashCode();
hc.add(42);
console.log(hc.toHashCode());
\`\`\`

| Bundle | Size | Temporal/Decimal refs |
|---|---|---|
| Before (no flag) | **175.1 KB** | yes |
| After (\`sideEffects: false\`) | **1.96 KB** | none |

**88× smaller.** The 175 KB figure exactly matches the size reported on \`sample-counter-with-dom-view\`.

## Why the flag is sound

The runtime is hand-authored and entirely declarative — function declarations, class declarations, const literals, no IIFEs, no top-level side-effectful registers. Even \`Symbol(\"delegate_listeners\")\` in \`delegates.ts\` is a single allocation read by the helpers in the same module; once the module is unused, the entire definition (including the symbol) is removed.

## Out of scope

\`sample-counter\` Vite build is currently broken on a separate, pre-existing issue: \`src/sample-counter/counter.ts(4,15) error TS1294: This syntax is not allowed when 'erasableSyntaxOnly' is enabled\` (parameter property in constructor). I did not measure the bundle through that path because the build does not complete. The minimal-entry \`bun build\` measurement above directly exercises the tree-shake path that consumers' bundlers will follow.

## Test plan

- [x] Bun runtime tests: 275/275 pass — flag has no runtime semantics, only affects bundler hints.
- [x] Bundle measurement before/after captured in the table above.

Closes #137

### DIFF
--- a/targets/js/metano-runtime/package.json
+++ b/targets/js/metano-runtime/package.json
@@ -4,6 +4,7 @@
   "type": "module",
   "module": "dist/index.js",
   "types": "dist/index.d.ts",
+  "sideEffects": false,
   "description": "Runtime support library for Metano-generated TypeScript code",
   "license": "MIT",
   "repository": {


danfma:
@gemini-code-assist thanks for the careful read. The flag is safe in this case — analyzed both ends of the import chain:

**1. metano-runtime's use of the polyfill is named-import only.** \`temporal-helpers.ts\` does \`import { Temporal } from \"@js-temporal/polyfill\"\` and references \`Temporal.PlainDate\` etc. through the local binding. There is no \`import \"@js-temporal/polyfill\"\` (bare import for global registration) anywhere in the runtime. Nothing in the runtime relies on \`globalThis.Temporal\` being patched.

**2. The polyfill itself declares targeted sideEffects.** \`@js-temporal/polyfill/package.json\` has:
\`\`\`json
\"sideEffects\": [\"./tsc-out/calendar.js\"]
\`\`\`
So when \`temporal-helpers.ts\` IS pulled (consumer calls \`dayNumber\`), the bundler honors that list and includes \`calendar.js\` plus everything else the named \`Temporal\` import transitively needs. When \`temporal-helpers.ts\` is NOT pulled (consumer only uses \`HashCode\` etc.), tree-shaking drops the import line and the polyfill is fully removed — the empirical 175.1 KB → 1.96 KB measurement confirms this.

So the runtime's \`sideEffects: false\` doesn't disturb the polyfill's own correctness — it only governs what survives in metano-runtime's own modules, which is genuinely declarative.

PR #142, opened 2026-04-27
fix(core): synthesize backing fields for captured primary-ctor params

### LINKED ISSUE #141: Primary constructor parameter captured by member methods is not emitted as field
## Problem

C# primary-constructor parameters that are referenced from outside the constructor body (i.e. captured by an instance method or another member) get an implicit backing field synthesized by Roslyn. The transpiler currently emits the captured parameter neither as a constructor parameter nor as a field, so any reference to it inside an instance method becomes an undefined identifier in the generated TypeScript.

## Reproduction

C# input:

\`\`\`csharp
using SampleCounterWithDomView.Models;
using SampleCounterWithDomView.Views;

namespace SampleCounterWithDomView.Presenters;

public sealed class CounterPresenter(ICounterView view, Counter initialState)
{
    private Counter _state = initialState;

    public void StartApplication(string containerName)
    {
        var renderer = new Renderer(containerName);
        renderer.Render(view);

        view.OnButtonClick = OnButtonClicked;
        view.ShowCounter(_state.Count);
    }

    private void OnButtonClicked()
    {
        _state = _state.Increment();
        view.ShowCounter(_state.Count);
    }
}
\`\`\`

Current TypeScript output:

\`\`\`ts
import type { Counter } from "#/models";
import { Renderer } from "#/views";

export class CounterPresenter {
  private _state: Counter;

  constructor(initialState: Counter) {
    this._state = initialState;
  }

  startApplication(containerName: string): void {
    const renderer = new Renderer(containerName);
    renderer.render(view);                          // BUG — `view` undefined
    view.onButtonClick = this.onButtonClicked.bind(this);
    view.showCounter(this._state.count);
  }

  private onButtonClicked(): void {
    this._state = this._state.increment();
    view.showCounter(this._state.count);            // BUG — `view` undefined
  }
}
\`\`\`

Symptoms:
1. \`view\` parameter is dropped from the emitted constructor signature.
2. The \`ICounterView\` import is missing entirely.
3. Member methods reference a bare \`view\` identifier that resolves to nothing — runtime ReferenceError.
4. The C# compiler internally synthesizes a backing field for \`view\` because it's captured by methods; the transpiler must do the same.

The \`initialState\` parameter is handled correctly because the user manually wrote \`private Counter _state = initialState;\` to capture it. Without that explicit assignment, \`view\` falls through.

## Expected

Either:

\`\`\`ts
import type { Counter } from "#/models";
import type { ICounterView } from "#/views";
import { Renderer } from "#/views";

export class CounterPresenter {
  private readonly _view: ICounterView;
  private _state: Counter;

  constructor(view: ICounterView, initialState: Counter) {
    this._view = view;
    this._state = initialState;
  }

  startApplication(containerName: string): void {
    const renderer = new Renderer(containerName);
    renderer.render(this._view);
    this._view.onButtonClick = this.onButtonClicked.bind(this);
    this._view.showCounter(this._state.count);
  }

  private onButtonClicked(): void {
    this._state = this._state.increment();
    this._view.showCounter(this._state.count);
  }
}
\`\`\`

Or the parameter-property form:

\`\`\`ts
constructor(private readonly view: ICounterView, initialState: Counter) {
  this._state = initialState;
}
\`\`\`

Either shape works as long as captured primary-ctor parameters survive on the instance.

## Detection rule

For each primary-constructor parameter:
1. If the parameter is referenced anywhere outside the constructor body (`SemanticModel` symbol-uses lookup against an `IParameterSymbol` whose `ContainingSymbol` is the primary ctor), Roslyn synthesizes a backing field. Roslyn exposes this via `IParameterSymbol.IsPrimaryConstructorParameter` plus a usage scan on the type's other members.
2. The transpiler must:
   - Add the parameter to the emitted constructor signature.
   - Emit a private (readonly when not reassigned) field on the class for the captured parameter.
   - Rewrite bare references to the parameter inside instance members as `this._parameterName` (or the chosen field name).
   - Add the parameter type to the file's imports.

## Acceptance

- [ ] Captured primary-ctor parameters are emitted as private fields and assigned in the constructor body.
- [ ] Bare references inside instance members rewrite to `this._param`.
- [ ] The parameter type is imported when it lives in another file/package.
- [ ] An explicit user-authored capture (\`private T _x = paramName;\`) still works as today — the auto-capture should not double-emit.
- [ ] A primary-ctor parameter that is **only** used inside the ctor body (e.g. forwarded to a base call) does not become a field.
- [ ] Static methods cannot capture — only instance members trigger the auto-field.
- [ ] Tests:
  - Plain captured parameter (`view` case above).
  - Mixed: one captured, one explicitly assigned to a field.
  - Captured + assigned (parameter assigned to a field AND referenced bare elsewhere) — should not double-emit.
  - Parameter shadowed by an instance field of the same name — disambiguate.
  - Multiple instance methods capturing the same parameter.

## Where to look

- \`src/Metano.Compiler/Extraction/IrClassExtractor.cs\` — primary-ctor parameter handling, where captured params would need detection.
- \`src/Metano.Compiler/Extraction/IrExpressionExtractor.cs\` — identifier resolution that would need to rewrite bare \`view\` references to \`this._view\`.
- \`src/Metano.Compiler.TypeScript/Bridge/IrToTsClassEmitter.cs\` — class emission, where new fields would land.
- \`src/Metano.Compiler.TypeScript/Transformation/ImportCollector.cs\` — needs to follow the captured parameter type for cross-package imports.

## Related

Surfaced by the SampleCounterWithDomView consumer (same sample that already drove issues #132 and #137).

danfma:
## Summary

C# 12 primary-constructor parameters that are referenced from a member body get an implicit backing field synthesized by Roslyn. The transpiler used to drop those parameters from the emitted constructor and leave bare identifier references inside member bodies, producing TypeScript that fails at runtime with \`ReferenceError\`.

Repro from the SampleCounterWithDomView consumer:

\`\`\`csharp
public sealed class CounterPresenter(ICounterView view, Counter initialState)
{
    private Counter _state = initialState;

    public void StartApplication(string container)
    {
        view.OnButtonClick = OnButtonClicked;  // captured
    }

    private void OnButtonClicked()
    {
        view.ShowCounter(_state.Count);  // bare \`view\` → ReferenceError
    }
}
\`\`\`

Before: \`view\` dropped from ctor signature, missing import, bare \`view\` references.
After: synthesized \`private readonly _view: ICounterView\` field + \`this._view = view;\` in ctor + \`this._view\` rewrite throughout member bodies.

## Mechanism

- New \`ImplicitCaptureDetector\` walks the type's syntax once before member extraction. For each \`IdentifierNameSyntax\` outside the primary-ctor's parameter list and any base-args span, it semantically resolves the symbol; when it matches a primary-ctor parameter the param is queued for capture. Parameters already captured by an explicit field initializer (\`private readonly Foo _foo = foo;\`) are skipped — the existing \`AnnotateCapturedParams\` pass already wires those end-to-end.
- The detector synthesizes a \`_paramCamelCase\` private readonly \`IrFieldDeclaration\` per implicit capture (\`IsCapturedByCtor: true\`). The backend's existing constructor-body emitter (\`IrToTsClassBridge.BuildSimpleConstructor\`) already consumes \`IsCapturedByCtor\` + \`IrConstructorParameter.CapturedFieldName\` to inject \`this._field = param;\` at the top of the ctor body.
- A new \`AnnotateImplicitCaptures\` pass surfaces the capture map on \`IrConstructorParameter.CapturedFieldName\`.
- \`IrExpressionExtractor\` adds an \`AsyncLocal<IReadOnlyDictionary<ISymbol, string>?>\` slot. \`IrClassExtractor.Extract\` sets it around \`ExtractMembers\`. \`ExtractIdentifierName\` rewrites resolved \`IParameterSymbol\` references to \`IrMemberAccess(IrThisExpression, fieldName)\`.

The rewrite is keyed on **symbol identity** (not name matching), so a method-body local that shadows the captured param remains a local — Roslyn's semantic resolution returns the local symbol, not the param.

## Trade-offs

- **Field-name collision** (a class with both a \`view\` primary-ctor param AND an existing \`_view\` member field): synthesis is skipped silently. The user's hand-written field stays intact; the unrewritten bare param reference surfaces as a TS compile error at consumption time. Surfacing a dedicated \`MS00xx\` diagnostic requires threading a diagnostic sink through the extraction pipeline — tracked as a follow-up.

## Test plan

- [x] 7 new tests in \`PrimaryConstructorCaptureTests\`:
  - Captured param synthesizes private readonly field + \`this._field\` accesses.
  - Explicitly-captured param keeps the user's field (no duplicate).
  - Mixed: explicit + implicit captures coexist.
  - Ctor-only param does not synthesize.
  - Multiple methods sharing the same captured param see one synthesized field.
  - Field-name collision skips synthesis and preserves the user's field.
  - Local shadowing the captured param does NOT get rewritten.
- [x] Full .NET suite: 779/779 pass.
- [x] csharpier-format clean.

- compiler-man: flagged Major (field-name collision, no detection) → fixed by silent-skip + follow-up note. Minor missing-tests → added shadow + collision coverage.
- bob: skipped this round; helpers are single-purpose, blank-line discipline applied. Will run on review feedback.

Closes #141

### DIFF
--- a/.claude/agent-memory/compiler-man/MEMORY.md
+++ b/.claude/agent-memory/compiler-man/MEMORY.md
@@ -1,3 +1,3 @@
 # compiler-man Agent Memory
 
-(Empty — will be populated as compiler-man accumulates project-specific findings and user feedback.)
+- [IR captured ctor params](ir_captured_ctor_params.md) — `CapturedFieldName` + `IsCapturedByCtor` already exist; extend `AnnotateCapturedParams`, don't redesign IR.
--- a/.claude/agent-memory/compiler-man/ir_captured_ctor_params.md
+++ b/.claude/agent-memory/compiler-man/ir_captured_ctor_params.md
@@ -1,0 +1,16 @@
+---
+name: IR captures ctor-param-to-field through CapturedFieldName
+description: How the IR models DI-style primary-ctor parameter capture into a backing field; the contract bridges already implement.
+type: project
+---
+
+The IR already carries the contract for "primary-ctor parameter captured into a backing field":
+
+- `IrConstructorParameter.CapturedFieldName` (string?, in `IR/IrConstructorDeclaration.cs`) names the field that holds the captured param.
+- `IrFieldDeclaration.IsCapturedByCtor` (bool, in `IR/IrMemberDeclaration.cs`) signals to bridges that the field's `Initializer` must be suppressed (the ctor body assigns it instead).
+- `IrClassExtractor.AnnotateCapturedParams` (post-pass after member + ctor extraction) is the single producer of these flags. Today it only handles **explicit** captures (`private readonly Foo _foo = foo;`) by matching field initializers that are bare `IrIdentifier`s pointing at a ctor parameter.
+- `IrToTsClassBridge.BuildCapturedCtorParams` and `BuildSimpleConstructor` already consume the contract on the TS side: captured params render with `TsAccessibility.None` (no shorthand-property promotion) and the ctor body emits `this._foo = foo;` per captured pair.
+
+**Why:** When extending capture detection (e.g. issue #141 — implicit captures via method-body references), do not invent new IR fields. Extend `AnnotateCapturedParams` to also synthesize fields for params that are referenced from non-ctor member bodies, and add a body-rewrite branch in `IrExpressionExtractor.ExtractIdentifierName` that turns a captured `IParameterSymbol` into `IrMemberAccess(this, fieldName)`. Bridges and Dart parity follow for free.
+
+**How to apply:** Before designing capture-related changes, check whether `CapturedFieldName`/`IsCapturedByCtor` already covers the shape. The bridge contract is the source of truth — match it, don't fork it.
--- a/src/Metano.Compiler/Extraction/ImplicitCaptureDetector.cs
+++ b/src/Metano.Compiler/Extraction/ImplicitCaptureDetector.cs
@@ -61,29 +61,53 @@ internal static class ImplicitCaptureDetector
             SymbolEqualityComparer.Default
         );
         foreach (var param in primaryCtor.Parameters)
+        {
+            // Parameters that match a public property of the type are
+            // already promoted to that property by C# 12's positional-
+            // record / primary-ctor surface; the existing constructor
+            // bridge handles them end-to-end. Synthesizing a private
+            // backing field would duplicate state and surface as
+            // double-emitted fields in regenerated samples.
+            if (IsPromotedToProperty(param, type))
+                continue;
             primaryParams[param] = param;
+        }
+
+        if (primaryParams.Count == 0)
+            return Empty;
 
         // Parameters captured by an explicit field initializer get their
         // synthesis suppressed early — the existing AnnotateCapturedParams
-        // pass already wires the field/param pair end-to-end. The guard
-        // also stops us from misclassifying the initializer's identifier
-        // (which lives inside the type's syntax tree) as an implicit
-        // capture.
-        var explicitlyCapturedParams = CollectExplicitlyCapturedParams(type, primaryParams);
+        // pass already wires the field/param pair end-to-end. Symbol-
+        // identity match guards against the false positive where an
+        // initializer references a static / base member that happens to
+        // share the parameter's name.
+        var explicitlyCapturedParams = CollectExplicitlyCapturedParams(
+            type,
+            compilation,
+            primaryParams
+        );
 
         var primaryCtorSyntaxNodes = CollectPrimaryCtorSyntaxNodes(type);
-        var existingFieldNames = type.GetMembers()
-            .OfType<IFieldSymbol>()
-            .Select(f => f.Name)
+        // TypeScript shares one namespace across fields, properties and
+        // methods, so the collision check has to consider every
+        // member's name — not just fields — or the synthesizer would
+        // happily emit a private field that shadows a public method.
+        var existingMemberNames = type.GetMembers()
+            .Select(m => m.Name)
             .ToHashSet(StringComparer.Ordinal);
         var captured = new HashSet<ISymbol>(SymbolEqualityComparer.Default);
 
-        foreach (var declaringRef in type.DeclaringSyntaxReferences)
+        // Walk only the members of THIS type (skip nested types); C# 12
+        // primary-ctor parameters are not in scope for a nested type,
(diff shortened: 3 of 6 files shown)


danfma:
@gemini-code-assist @copilot-pull-request-reviewer @chatgpt-codex-connector — all findings addressed across d986dd6 + 10573df. Build now green.

**d986dd6** (build fix + first review pass):
- **Build break (samples regen)** — root cause: detector synthesized a backing field for params that were already promoted to a public property (positional records, parameter-properties). Added `IsPromotedToProperty` skip with case-insensitive name match, mirroring \`IrConstructorExtractor.ResolvePromotionInfo\`. Sample regen byte-identical.
- **Gemini high (CollectExplicitlyCapturedParams)** — switched from name-based match to \`SemanticModel.GetSymbolInfo\` symbol-identity. Param signature now takes \`Compilation\`.
- **Gemini medium (collision check)** — \`existingMemberNames\` now considers every member (fields, properties, methods) instead of fields alone.
- **Gemini medium (nested-type walk)** — replaced \`typeSyntax.DescendantNodes()\` with \`EnumerateMemberSyntaxNodes\` that iterates only direct members of the type and skips \`INamedTypeSymbol\` children. Primary-ctor params are not in scope for nested types.

**10573df** (second review pass):
- **Copilot (initializer-rewrite ordering)** — added \`IsInsideInitializer\` walk in \`IrExpressionExtractor.ExtractIdentifierName\`. Field/property initializers keep the bare param reference; the rewrite to \`this._field\` only fires inside method bodies / accessors. Fixes the runtime \`undefined\` that would surface on \`private readonly int _x = view.Count\` because field init runs before the ctor body's \`this._view = view;\`.
- **Codex P2 + Copilot (TextSpan across syntax trees)** — \`IsInsidePrimaryCtor\` now matches both \`SyntaxTree\` identity and span. Partial types declared across multiple files no longer spuriously match a primary-ctor scope from a sibling file.

**Codex P1 (exclude initializers from scanning)** intentionally not applied — Roslyn synthesizes a backing field whenever the param is referenced anywhere in the type, including initializers, so the SCAN must count those references too. The ORDERING bug Codex hinted at is what Copilot's finding addresses head-on; suppressing the rewrite at extraction time covers both.

781/781 tests pass (added \`PromotedProperty_DoesNotSynthesizeBackingField\` and \`CapturedParam_ReadFromFieldInitializer_KeepsBareReference\`). Sample regen produces zero diff.

PR #143, opened 2026-04-28
fix(core): evaluate impure receiver once on method-group .bind

### LINKED ISSUE #133: Method group .bind(receiver) re-evaluates non-pure receivers

danfma:
## Summary

Method-group → delegate bind (PR #134) emits \`obj.method.bind(obj)\` — the receiver expression appears twice. For pure receivers (\`this\`, identifiers, member chains over those) duplication is observably equivalent. For receivers with side effects (calls, indexers, mutating getters) the chain runs twice.

\`\`\`csharp
target.OnTick = GetCache().Service.Tick;
\`\`\`

Before:
\`\`\`ts
target.onTick = this.getCache().service.tick.bind(this.getCache().service);
//              ^^^^^^^^^^^^^^^^^^^^^^^^                ^^^^^^^^^^^^^^^^^ second eval
\`\`\`

After:
\`\`\`ts
target.onTick = ((__r: any) => __r.tick.bind(__r))(this.getCache().service);
\`\`\`

## Mechanism

\`BuildBoundReference(IrMemberAccess instanceAccess)\` checks the receiver via the existing \`IsSimpleReceiver\` predicate. Pure shape → \`obj.method.bind(obj)\`. Impure shape → \`IrCallExpression(IrLambdaExpression([__r], return __r.method.bind(__r)), [originalReceiver])\`. The IIFE captures the receiver once.

## Test plan

- [x] \`MethodGroupAssignment_ImpureReceiver_EvaluatesChainOnce\` — asserts the IIFE shape and counts \`getCache()\` occurrences (expected 1).
- [x] Pre-existing pure-receiver tests still pass — no IIFE overhead when not needed.
- [x] Full suite: 782/782 pass.

Closes #133

### DIFF
--- a/src/Metano.Compiler/Extraction/IrExpressionExtractor.cs
+++ b/src/Metano.Compiler/Extraction/IrExpressionExtractor.cs
@@ -1288,21 +1288,34 @@ public sealed class IrExpressionExtractor
     /// method's body still runs against the current instance, which
     /// matches the C# semantics of <c>base</c>-qualified method
     /// groups (the dispatch is non-virtual but the receiver is
-    /// still <c>this</c>).
+    /// still <c>this</c>). The substitution recurses through
+    /// <see cref="IrMemberAccess"/> towers so a chain rooted in
+    /// <c>base</c> (<c>base.Property.Method</c>) becomes
+    /// <c>this.property.method.bind(this.property)</c> instead of
+    /// the invalid <c>super.property</c>.
     /// </summary>
     private static IrExpression BindArgumentFor(IrExpression receiver) =>
-        receiver is IrBaseExpression ? new IrThisExpression() : receiver;
+        receiver switch
+        {
+            IrBaseExpression => new IrThisExpression(),
+            IrMemberAccess access => access with { Target = BindArgumentFor(access.Target) },
+            _ => receiver,
+        };
 
     /// <summary>
     /// Builds the <c>.bind(receiver)</c> expression for an instance
-    /// method group. When the receiver chain is pure (identifier,
-    /// <c>this</c>, <c>base</c>, or a member-access tower over those)
-    /// emits the simple <c>obj.method.bind(obj)</c> shape. Otherwise
-    /// the receiver has observable side effects (call expressions,
-    /// indexers, getters that mutate) and duplicating it would run
-    /// the side effect twice — wrap the bind site in an IIFE arrow
-    /// that captures the receiver in a temporary so the chain is
-    /// evaluated exactly once:
+    /// method group. When the receiver chain reduces to a pure IR
+    /// shape (identifier, <c>this</c>, <c>base</c>, or a chain of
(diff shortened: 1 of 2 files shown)


danfma:
@gemini-code-assist @copilot-pull-request-reviewer addressed in 3184016.

- **Gemini high (\`BindArgumentFor\` not recursive)** — \`base.Property.Method\` chains now walk through \`IrMemberAccess\` and substitute \`this\` for the \`super\` root anywhere in the chain. Locked in by \`BaseMemberAccessChain_MethodGroup_RecursivelySubstitutesSuper\`.
- **Copilot (doc inaccuracy)** — reworded \`BuildBoundReference\` doc to match the shape-based \`IsSimpleReceiver\` predicate. Symbol-aware purity tracking (catching property getters with side effects) is flagged as a separate follow-up — would tighten predicate used by 3 unrelated code paths.

**Gemini medium (\`IsSimpleReceiver\` over-permissive for property chains)** intentionally not applied: the predicate is shared with null-conditional assignment lowering and the \`[This]\` delegate path. Tightening would force IIFE/temp-binding on every member-chain access in those paths too — pre-existing concern that needs symbol-info plumbing to do safely. Tracked as future work alongside the Copilot doc note.

783/783 tests pass.

PR #144, opened 2026-04-28
feat(ts): emit extends + abstract for class inheritance

### LINKED ISSUE #118: Class inheritance: emit extends for class A : B
## Motivation

A C# class that extends another transpilable class should produce a TypeScript \`class A extends B\` declaration and forward constructor / member lookups through the prototype chain.

```csharp
[Transpile]
public abstract class Shape
{
    public abstract double Area();
}

[Transpile]
public sealed class Circle(double radius) : Shape
{
    public override double Area() => Math.PI * radius * radius;
}
```

Expected:

```ts
export abstract class Shape {
  abstract area(): number;
}

export class Circle extends Shape {
  constructor(private readonly radius: number) { super(); }
  area(): number { return Math.PI * this.radius * this.radius; }
}
```

\`IrClassExtractor.cs:29-30\` already captures the base type. The TS bridge needs to route every class shape (plain, record-adjacent, overload dispatcher) through the \`extends\` clause and inject the \`super(...)\` call into the lowered constructor.

## Proposal

- \`IrToTsClassBridge\` emits \`extends BaseName\` when \`IrClass.BaseType\` is set to a meaningful base (i.e., not \`object\`).
- \`IrConstructorExtractor\` already tracks \`: base(...)\` calls; bridge those into a TypeScript \`super(...)\` call at the top of every constructor body.
- When the constructor list is synthesized (no explicit ctors in C# but a base is present), generate a single \`constructor(...args) { super(...args); }\` or the minimal shape to satisfy TypeScript's \"derived class must call super\".
- Method overrides emit without the \`override\` keyword (TypeScript infers it), keeping the existing naming / accessibility policy.

## Semantics

- **Single base class.** \`class A : B\` → \`class A extends B\`.
- **Base class with primary constructor args.** \`class A(int x) : B(x)\` → TS constructor forwards via \`super(x)\` (ties into the primary-ctor capture issue).
- **Abstract base.** Base keeps \`abstract\` modifier; derived removes it per member.
- **\`object\` base.** Silent — no \`extends\` clause.
- **External / \`[NoEmit]\` base.** Emits \`extends\` with the declared name but skips import emission (ambient type), matching current external-type handling.

## Open questions

- **Cross-package base class.** Import collector must add \`import { Base } from \"pkg/path\"\`; confirm existing cross-project infra picks it up.
- **Base with generics.** \`class A : List<int>\` → \`class A extends List<number>\` using the BCL mapping. Agree on policy when the base is a BCL type that maps to a non-class (\`IEnumerable\` → \`Iterable\` which cannot be \`extends\`'d in TS). Likely diagnose.
- **Constructor overload dispatcher.** When the derived class has multiple constructors, the dispatcher must still emit a single \`super(...)\` path per branch.

## Acceptance

- [ ] \`IrToTsClassBridge\` emits \`extends BaseName\` for every class with a meaningful base.
- [ ] Every constructor body emits the correct \`super(...)\` call, including synthesized ones.
- [ ] Primary-ctor capture and base forwarding compose cleanly (see primary-ctor capture issue).
- [ ] Abstract modifier is honored on base vs derived.
- [ ] Cross-package base classes resolve via the import collector.
- [ ] Tests:
  - Plain single-base lowering.
  - Primary ctor with base args (\`class A(int x) : B(x)\`).
  - Abstract base with override.
  - Cross-package base.
  - Diagnostic when the base is an uninhabitable TS type (e.g., mapped BCL type that is not a class).

danfma:
## Summary

Wires the existing class-inheritance infra to cover plain (non-record) classes per #118.

\`\`\`csharp
[Transpile]
public abstract class Shape
{
    public abstract double Area();
}

[Transpile]
public class Circle(double radius) : Shape
{
    public override double Area() => 3.14159 * radius * radius;
}
\`\`\`

emits

\`\`\`ts
export abstract class Shape {
  abstract area(): number;
}

export class Circle extends Shape {
  constructor(readonly radius: number) {
    super();
  }
  area(): number { return 3.14159 * this.radius * this.radius; }
}
\`\`\`

## Mechanism

- \`BuildSimpleConstructor\` emits \`super(...)\` whenever the class extends a base, even when the base ctor takes no arguments — TS rejects derived classes that read \`this\` without calling super. Previously the empty-args case skipped the call entirely.
- \`TsClass\` and \`TsMethodMember\` carry a new \`Abstract\` flag; the printer emits the \`abstract\` modifier on both. The TS class emitter propagates \`IrTypeSemantics.IsAbstract\` to the class and \`IrMethodSemantics.IsAbstract\` to each method.
- \`BuildMethod\` skips the not-implemented stub for abstract methods so the printer emits \`area(): number;\` instead of a bogus body.

The pre-existing \`BuildExtends\`, \`ResolveSuperArgs\`, primary-ctor capture, and constructor-overload-dispatcher paths already covered records, base-args forwarding, cross-package imports, and the non-transpilable / object-base filters.

## Test plan

- [x] 6 new tests in \`ClassInheritanceTests\`: plain class extends, abstract base + abstract method, primary ctor with base args, explicit ctor with \`: base(...)\`, non-transpilable base skips extends, \`object\` base skips extends.
- [x] Existing \`InheritanceTranspileTests\` (records) all pass — \`super(...)\` shape unchanged for non-empty arg lists.
- [x] Sample regen (SampleTodo, SampleIssueTracker, SampleTodo.Service) produced zero diff.
- [x] Full .NET suite: 789/789 pass.
- [x] csharpier-format clean.

Closes #118

### DIFF
--- a/src/Metano.Compiler.TypeScript/Bridge/IrToTsClassBridge.cs
+++ b/src/Metano.Compiler.TypeScript/Bridge/IrToTsClassBridge.cs
@@ -253,7 +253,12 @@ internal static class IrToTsClassBridge
     )
     {
         var name = IrToTsNamingPolicy.ToMethodName(method.Name, method.Attributes);
-        var body = IrToTsBodyHelpers.LowerOrNotImplemented(method.Body, method.Name, bclRegistry);
+        // Abstract methods carry no body in C#; the printer emits the
+        // signature followed by `;` and skips the not-implemented stub
+        // that would otherwise replace a missing body.
+        IReadOnlyList<TsStatement> body = method.Semantics.IsAbstract
+            ? []
+            : IrToTsBodyHelpers.LowerOrNotImplemented(method.Body, method.Name, bclRegistry);
         return new TsMethodMember(
             name,
             parameters,
@@ -263,7 +268,8 @@ internal static class IrToTsClassBridge
             Async: method.Semantics.IsAsync,
             Generator: method.Semantics.IsGenerator,
             Accessibility: MapAccessibility(method.Visibility),
-            TypeParameters: typeParameters
+            TypeParameters: typeParameters,
+            Abstract: method.Semantics.IsAbstract
         );
     }
 
@@ -767,7 +773,13 @@ internal static class IrToTsClassBridge
     )
     {
         var body = new List<TsStatement>();
-        if (superArgs is not null && superArgs.Count > 0)
+        // `superArgs is not null` signals that the class extends a
+        // base — TypeScript requires every derived constructor to
+        // call `super(...)` before reading `this`, even when the
+        // base constructor takes no arguments. An empty list lowers
+        // to a bare `super()` call rather than skipping the
+        // statement.
+        if (superArgs is not null)
         {
             body.Add(
                 new TsExpressionStatement(
--- a/src/Metano.Compiler.TypeScript/Bridge/IrToTsClassEmitter.cs
+++ b/src/Metano.Compiler.TypeScript/Bridge/IrToTsClassEmitter.cs
@@ -218,7 +218,8 @@ public sealed class IrToTsClassEmitter(TypeScriptTransformContext context)
                 classMembers,
                 Extends: extendsType,
                 Implements: IrToTsClassBridge.BuildImplements(ir),
-                TypeParameters: IrToTsClassBridge.BuildTypeParameters(ir)
+                TypeParameters: IrToTsClassBridge.BuildTypeParameters(ir),
+                Abstract: ir.Semantics.IsAbstract
             )
         );
     }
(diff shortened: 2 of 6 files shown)


danfma:
@chatgpt-codex-connector @copilot-pull-request-reviewer addressed in 0cd31cd.

- **Codex P1**: real bug — record synthesis emits \`with(...) { return new T(...) }\` which TS rejects on \`abstract class\`. Suppress the abstract modifier when \`ir.Semantics.IsRecord\` is true. Locked in by \`AbstractRecord_SuppressesAbstractModifier\`. Constructor-aware \`with\` synthesis is a follow-up.
- **Copilot**: reworded the test comment — the super-call requirement reads as a type-checker rule rather than runtime-only.

790/790 tests pass.

PR #147, opened 2026-04-28
feat(ts): emit default values for method and function parameters

### LINKED ISSUE #115: Emit default values for method and function parameters
## Motivation

C# parameters can carry explicit default values:

```csharp
public HtmlElement Create(string tagName, bool attached = true)
```

The IR already captures the default expression (`IrParameter.HasDefaultValue` + `IrParameter.DefaultValue`, see `src/Metano.Compiler/Extraction/IrMethodExtractor.cs:32-33` and the matching paths in `IrModuleFunctionExtractor`, `IrConstructorExtractor`), but only two emission paths consume it:

- **Constructor parameters** (`TsConstructorParam.DefaultValue`) print the `= expr` suffix via `Printer.cs:596`.
- **`[PlainObject]` fields** mark the field optional when the constructor parameter has a default (`IrToTsPlainObjectBridge.cs:49`).

Every other parameter site in the TypeScript target drops the default silently because `TsParameter` does not carry a `DefaultValue` slot. The generated methods/functions accept fewer call shapes than the source intends.

Example — current output today:

```csharp
[Transpile]
public static class Builder
{
    public static string Greet(string name, string greeting = "hello") => $"{greeting}, {name}";
}
```

```ts
// builder.ts (generated)
export function greet(name: string, greeting: string): string {
  return `${greeting}, ${name}`;
}
```

Callers that relied on the default now have to pass the argument explicitly in TS, defeating the ergonomic purpose of the C# default.

## Proposal

Thread the captured default expression through the TypeScript emission so every parameter-bearing surface renders it:

- Extend `TsParameter` with `TsExpression? DefaultValue` and `bool Optional`.
- Update every `new TsParameter(...)` site in `src/Metano.Compiler.TypeScript/Bridge/` to forward `p.DefaultValue` (lowered via `IrToTsExpressionBridge`) and `p.HasDefaultValue` to the TS parameter.
- Update `Printer.PrintTsParameter` to emit the default suffix the same way constructor parameters do today. The `= expr` form implies optional, so we never emit `?:` and `=` simultaneously.
- Interface / abstract / overload-signature slots (declaration-only positions where TS forbids `= value`) continue to lower as `name?: Type` using `Optional`.

### Concrete methods and module functions

```csharp
public static string Greet(string name, string greeting = "hello") => ...
```

```ts
export function greet(name: string, greeting: string = "hello"): string { ... }
```

### Interface methods / abstract signatures

TypeScript disallows `= value` in declaration positions. Lower to optional instead:

```csharp
public interface IFormatter
{
    string Format(string value, bool upper = false);
}
```

```ts
export interface IFormatter {
  format(value: string, upper?: boolean): string;
}
```

### Overload dispatchers

`IrToTsOverloadDispatcherBridge` generates per-branch signatures that already need the optional hint; the default expression can be carried as well when the branch is the single concrete implementation. When multiple overloads collapse into one, fall back to `Optional` because the runtime body needs to check `arg === undefined` anyway.

### Constructor parameters

Already covered via `TsConstructorParam.DefaultValue`. Left as-is for backward compatibility; the new `TsParameter` slot does not duplicate that pathway.

### PlainObject records

Already mapped — default-valued constructor parameters lower to optional fields (see `IrToTsPlainObjectBridge.cs:49`). No change.

### Lambdas

Roslyn does not expose default values on lambda parameters, so nothing to emit. Validator leaves them untouched.

## Emission rules summary

| Position                     | TS form                    |
|------------------------------|----------------------------|
| Concrete function/method     | `name: T = value`          |
| Abstract/interface method    | `name?: T`                 |
| Overload branch (concrete)   | `name: T = value`          |
| Overload dispatcher union    | `name?: T` (runtime check) |
| Constructor parameter        | `name: T = value` (today)  |
| `[PlainObject]` field        | `readonly name?: T`        |

## Open questions

- **Complex defaults.** `IrMethodExtractor.ExtractDefaultValue` already lowers literal defaults. Enum/member-access defaults (`Priority.Normal`) should round-trip as TS identifiers — check whether the current extractor covers that; if not, extend.
- **Cross-assembly.** Defaults on referenced-assembly methods must resolve through the referenced compilation, matching how `[Inline]` now traverses `CompilationReference`.
- **`default(T)` literal.** Maps to `undefined` today for most reference types; agree on whether we should instead emit the TS equivalent (`null` vs `undefined`) or refuse to lower with `MS00XX UnsupportedDefault`.
- **Optional-only callers in the TS consumer.** Emitting `= value` is a TS declaration; consumers that call the function with fewer args still get the runtime default. Confirm the sample suites (`sample-todo`, `sample-issue-tracker`) do not regress.

## Acceptance

- [ ] `TsParameter` gains `TsExpression? DefaultValue` and `bool Optional`.
- [ ] All `new TsParameter(...)` call sites in `src/Metano.Compiler.TypeScript/Bridge/` forward the IR default through `IrToTsExpressionBridge`.
- [ ] `Printer.PrintTsParameter` emits `= expr` for concrete positions and `?:` for declaration-only positions, mirroring existing constructor behavior.
- [ ] Overload dispatcher honors defaults on the concrete branch and falls back to `Optional` in the public union signature.
- [ ] Interface methods emit `name?: T` instead of dropping the default.
- [ ] `default(T)` and enum-member defaults (`Priority.Normal`) round-trip as readable TS expressions.
- [ ] Cross-assembly defaults resolve through referenced compilations (parallel to the `[Inline]` path).
- [ ] Tests:
  - Module function with literal default.
  - Instance method with literal default.
  - Interface method with default (emits `?:`).
  - Enum-member default (`= Priority.Normal`).
  - Overload group with one concrete implementation keeping `= value`.
  - Cross-assembly method default lowering.
  - Sample regression: `sample-todo`, `sample-issue-tracker`, `sample-counter-with-dom-view` stay green with defaults preserved end-to-end.

Related: `src/Metano.Compiler/IR/IrParameter.cs`, `src/Metano.Compiler.TypeScript/TypeScript/AST/TsParameter.cs`, `src/Metano.Compiler.TypeScript/TypeScript/Printer.cs`, `src/Metano.Compiler.TypeScript/Bridge/IrToTsPlainObjectBridge.cs`.

danfma:
## Summary

C# parameter defaults (\`Greet(string name, string greeting = \"hello\")\`) now propagate to the emitted TypeScript signature so callers preserve the optional ergonomics the source intends. Closes #115.

\`\`\`csharp
public string Greet(string name, string greeting = \"hello\") => $\"{greeting}, {name}\";
\`\`\`

emits

\`\`\`ts
greet(name: string, greeting: string = \"hello\"): string {
  return \`\${greeting}, \${name}\`;
}
\`\`\`

## Mechanism

- \`TsParameter\` gains a \`DefaultValue: TsExpression?\` field. The printer emits \`name: T = expr\`, biased over the optional \`?\` suffix when both are set (TS forbids combining them and the call site needs the seeded value, not \`undefined\`).
- \`IrToTsClassEmitter.LowerMethodSignature\` and \`IrToTsModuleBridge.ConvertFunction\` thread \`HasDefaultValue\` + \`DefaultValue\` from the IR through the standard expression bridge — BCL-mapped initializers (decimal literals, enum members) get their target form.

Declaration-only positions (interface signatures, abstract method declarations, overload signatures) are not touched here — they keep using the \`?\` suffix because TS forbids initializers in those slots. Tracked as a follow-up if any of those surfaces start carrying defaults in real samples.

## Test plan

- [x] 5 new tests in \`DefaultParameterValueTests\`: instance method with string default, module function with bool default, numeric default, null default on a nullable, no-default parameter stays required.
- [x] Sample regen (SampleTodo, SampleIssueTracker) byte-identical — current samples don't use parameter defaults.
- [x] Full .NET suite: 795/795 pass.
- [x] csharpier-format clean.

### DIFF
--- a/src/Metano.Compiler.TypeScript/Bridge/IrToTsClassEmitter.cs
+++ b/src/Metano.Compiler.TypeScript/Bridge/IrToTsClassEmitter.cs
@@ -399,12 +399,17 @@ public sealed class IrToTsClassEmitter(TypeScriptTransformContext context)
         IReadOnlyList<TsTypeParameter> TypeParameters
     ) LowerMethodSignature(IrMethodDeclaration irMethod, IrClassDeclaration owner)
     {
+        // TypeScript forbids parameter initializers in declaration-
+        // only positions (abstract / overload / interface signatures).
+        // For abstract methods drop the default expression but keep
+        // the parameter optional so callers may still omit it.
+        var emitDefaults = !irMethod.Semantics.IsAbstract;
         var parameters = irMethod
             .Parameters.Select(p => new TsParameter(
                 IrToTsNamingPolicy.ToParameterName(p.Name),
                 IrToTsTypeMapper.Map(p.Type, BclOverrides),
-                Optional: p.IsOptional,
-                DefaultValue: LowerDefaultValue(p)
+                Optional: p.IsOptional || (!emitDefaults && p.HasDefaultValue),
+                DefaultValue: emitDefaults ? LowerDefaultValue(p) : null
             ))
             .ToList();
         var returnType = IrToTsTypeMapper.Map(irMethod.ReturnType, BclOverrides);
--- a/src/Metano.Compiler.TypeScript/Bridge/IrToTsClassEmitter.cs
+++ b/src/Metano.Compiler.TypeScript/Bridge/IrToTsClassEmitter.cs
@@ -402,7 +402,9 @@ public sealed class IrToTsClassEmitter(TypeScriptTransformContext context)
         var parameters = irMethod
             .Parameters.Select(p => new TsParameter(
                 IrToTsNamingPolicy.ToParameterName(p.Name),
-                IrToTsTypeMapper.Map(p.Type, BclOverrides)
+                IrToTsTypeMapper.Map(p.Type, BclOverrides),
+                Optional: p.IsOptional,
+                DefaultValue: LowerDefaultValue(p)
             ))
             .ToList();
         var returnType = IrToTsTypeMapper.Map(irMethod.ReturnType, BclOverrides);
@@ -619,6 +621,18 @@ public sealed class IrToTsClassEmitter(TypeScriptTransformContext context)
         IrToTsTypeMapper.Map(irType, BclOverrides);
 
     private BclExportTypeOverrides BclOverrides => _context.BclOverrides;
+
+    /// <summary>
+    /// Lowers an <see cref="IrParameter"/>'s default-value expression
+    /// into the TS form, or returns <c>null</c> when the parameter
+    /// has no explicit default. Routes through the standard
+    /// expression bridge so BCL-mapped initializers (decimal
+    /// literals, enum members, etc.) get their target form.
+    /// </summary>
+    private TsExpression? LowerDefaultValue(IrParameter parameter) =>
+        parameter.HasDefaultValue && parameter.DefaultValue is not null
+            ? IrToTsExpressionBridge.Map(parameter.DefaultValue, _context.DeclarativeMappings)
+            : null;
 
     /// <summary>
     /// Resolves the <c>super(...)</c> argument list for the bridge. Returns
--- a/src/Metano.Compiler.TypeScript/Bridge/IrToTsModuleBridge.cs
+++ b/src/Metano.Compiler.TypeScript/Bridge/IrToTsModuleBridge.cs
@@ -44,7 +44,11 @@ public static class IrToTsModuleBridge
         var parameters = fn
             .Parameters.Select(p => new TsParameter(
                 TypeScriptNaming.ToCamelCase(p.Name),
-                IrToTsTypeMapper.Map(p.Type)
+                IrToTsTypeMapper.Map(p.Type),
+                Optional: p.IsOptional,
+                DefaultValue: p.HasDefaultValue && p.DefaultValue is not null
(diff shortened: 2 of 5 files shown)


danfma:
@gemini-code-assist @copilot-pull-request-reviewer @chatgpt-codex-connector addressed in 873186e.

- **Abstract-method initializers (Gemini high, Codex P1)** — TS forbids \`= expr\` in abstract declarations. \`LowerMethodSignature\` now drops the initializer for abstract methods and promotes the parameter to optional (\`title?: string\`) so callers may still omit it. Locked in by \`AbstractMethod_DefaultValue_EmitsOptionalWithoutInitializer\`.
- **Lambda scope claim (Gemini medium, Copilot)** — reworded the test class summary. Lambdas are out of scope (C# disallows defaults on lambda parameters); declaration-only sites drop the initializer per the PR description's contract.
- **Brittle \`DoesNotContain(\"=\")\` (Copilot)** — replaced with a tighter \`Contains(\"echo(value: string): string {\")\` match on the signature.

**Cross-assembly defaults dropped (Copilot)** — pre-existing limitation: \`IrMethodExtractor.ExtractDefaultValue\` returns \`null\` when the declaring syntax is unreachable (referenced-assembly methods), so cross-assembly optional defaults still get dropped. Tracked as a follow-up — surfacing them needs a constant-value reader off the \`IParameterSymbol\` rather than a syntax round-trip.

796/796 tests pass.

Issue #148, opened 2026-04-28
ImportCollector misses type references inside function/delegate types


danfma:
## Problem

`ImportCollector.CollectFromType` does not recurse into `TsFunctionType`. When a method or delegate declares a parameter typed as a function (e.g. `Func<T, R>`, `Action<T>`, or a custom delegate) whose **parameters or return type** reference another transpilable type, that type's import is silently dropped from the emitted file.

## Reproducer

In a sample, declare:

```csharp
namespace Foo;

public interface IWidget { HtmlElement Build(); }

public delegate IWidget ViewFn<TState>(TState state, Action<TState> setState);
// (or equivalently, Func<TState, Action<TState>, IWidget>)

public static class App
{
    public static void Mount<TState>(string id, TState s0, ViewFn<TState> view) { /* ... */ }
}
```

Transpile. The emitted `app.ts` contains:

```ts
// no import { IWidget } from "./i-widget"
export class App {
  static mount<TState>(id: string, s0: TState, view: (state: TState, setState: (obj: TState) => void) => IWidget): void { /* ... */ }
}
```

`tsgo` rejects with:

```
src/foo/app.ts(N,M): error TS2304: Cannot find name 'IWidget'.
```

## Root cause

`src/Metano.Compiler.TypeScript/Transformation/ImportCollector.cs:668-716` — `CollectFromType` switches on `TsNamedType`, `TsArrayType`, `TsPromiseType`, `TsUnionType`, `TsIntersectionType`. There is **no case for `TsFunctionType`** (`src/Metano.Compiler.TypeScript/TypeScript/AST/TsFunctionType.cs:12`), so its `Parameters[].Type` and `ReturnType` are never visited.

## Suggested fix

Add a case to `CollectFromType`:

```csharp
case TsFunctionType func:
    if (func.ThisType is not null)
        CollectFromType(func.ThisType, names, crossPackageOrigins);
    foreach (var p in func.Parameters)
        CollectFromType(p.Type, names, crossPackageOrigins);
    CollectFromType(func.ReturnType, names, crossPackageOrigins);
    break;
```

## Workarounds in user code

Until fixed, avoid putting transpilable type references inside function-type parameters. Pass the type through a separate **method parameter** instead — those are visited correctly. Example: split the body so `IWidget` appears as a parameter of an internal helper method, which forces the import.

## Impact

[...]

PR #149, opened 2026-04-28
feat: add SampleCounterWithDomView, expose DOM APIs, and rewrite V3 as Mvu DSL

danfma:
## Summary

Three commits, building up DOM-targeted samples:

1. **`feat: add SampleCounterWithDomView and TypeScript DOM bindings`** — initial DOM bindings (Document, Element, Node, HtmlElement + button/div/span/body, EventTarget, Window) plus a counter sample wired through them.
2. **`feat: expose some DOM APIs through a new library`** — promotes the bindings into `bindings/Metano.TypeScript.DOM/`, splits the original into `SampleCounterV2` (imperative View+Presenter) and a placeholder `SampleCounterV3`, adds a typed `Document.CreateElement<T>(HtmlElementType.Of<T>)` extension, and tightens supporting compiler bits.
3. **`feat(samples): rewrite SampleCounterV3 as mini-Mvu/Flutter DSL`** — turns V3 into a Flutter-style DSL on top of the bindings: `IWidget`, `App.Mount`, `StateHolder`, and three widgets (`Text`, `Button`, `Column`). View is a state→widget function; `setState` re-renders the whole subtree (no diffing yet).

## Why V3 looks like this

Class-based widgets, but **interface dispatch** instead of `abstract Widget` — keeps the design portable to a future V4 once `abstract`/`virtual`/`override` ergonomics around `StatelessWidget`/`StatefulWidget` are settled. Stress-tests features the transpiler already handles well: sealed classes implementing an interface, generic delegate, lambda closure capture, collection expressions for children, primary-ctor field synthesis.

## Bug surfaced

`ImportCollector.CollectFromType` does not recurse into `TsFunctionType`, so a transpilable type referenced inside a `Func<>` / delegate signature is silently dropped from the emitted file's imports. Reported in #148 with diagnosis and a 4-line patch suggestion.

V3 works around it by routing the widget through a private `Apply(IWidget, HtmlElement)` helper — `IWidget` as a method parameter gets collected normally.

## Test plan

- [x] `dotnet build Metano.slnx -m:1` clean
- [x] `dotnet run --project tests/Metano.Tests/` — 797/797 pass
- [x] V3 transpiles (11 files) and `tsgo --noEmit -p tsconfig.app.json` clean
- [x] `tsgo -b` builds V3 to `lib/`
- [ ] `bun run --cwd targets/js/sample-counter-v3 dev` — counter increments on click in browser

Closes nothing (sample-only; tracked alongside #148 for the import bug).


danfma:
Thanks @chatgpt-codex-connector and @gemini-code-assist — review fixes pushed in 982a91e.

- **Non-static `[Inline]` methods** (Codex P1, Gemini medium on `CSharpSourceFrontend.cs`): rejected at validation time with an explanatory MS0016. Same shape as the existing static-property check. Test: `Inline_OnInstanceMethod_EmitsMs0016`.
- **Argument binding by parameter, not syntax order** (Codex P2, Gemini high on `IrExpressionExtractor.cs`): the expander now classifies caller args into a positional prefix + named map, walks the callee's parameters, and falls back to `BuildLiteralForDefault` when the caller skipped an optional. Tests: `Inline_Method_NamedArguments_BindsByParameterName`, `Inline_Method_OmittedOptional_FillsExplicitDefault`.
- **`Span`/`Button` missing `readonly`** (Gemini high on `HtmlElementType.cs`): both fields are now `static readonly`. Without that, MS0016 would have fired on the binding itself once anything imported it.
- **Duplicate `TryGetSemanticModel`** (Gemini medium): consolidated into `SymbolHelper.TryGetSemanticModel`. Both `CSharpSourceFrontend` and `IrExpressionExtractor` now go through the shared helper.

`dotnet run --project tests/Metano.Tests/`: 800/800 (3 new).

danfma:
Round 2 — pushed in 9836c4f. Thanks @copilot-pull-request-reviewer.

- **`[Inline]` XML doc out of date**: rewrote the doc on `InlineAttribute` to cover static methods (single-expression body, named/optional argument binding) and to remove the obsolete "methods are invalid" line.
- **XSS via `InnerHtml` in `Text` and `Button`**: added `[Name("textContent")] string? TextContent` to the `Element` binding. Both `SampleCounterV3` widgets now write to `TextContent` so plain labels can't smuggle markup. `InnerHtml` stays on the binding for explicit-trust use cases.
- **`holder.set.bind(holder)` allocated twice**: bound once into `Action<TState> setState = holder.Set;` in `App.Mount` and reused for the initial render and the `OnChange` callback. Generated TS now does a single `bind`.
- **`lib/` tracked alongside source**: untracked the committed `lib/` artifacts under `sample-counter-v2` and `sample-counter-v3`, added `lib` to each `.gitignore`. `tsgo -b` and `bun run dev` regenerate them locally.
- **`dev` port collision (V2 and V3 both on 3000)**: V2 → 3001, V3 → 3002. Root `bun run dev` can now spin both up in parallel.

V2's existing `InnerHtml = "0"` / `"Click me"` / `counter.ToString()` was not flagged and uses literal/numeric strings — left it alone to keep V2's "imperative DOM" angle untouched.

`dotnet run --project tests/Metano.Tests/`: 800/800. `tsgo -b` clean for both samples.

Issue #28, opened 2026-04-11
Dart target (`Metano.Compiler.Dart`)


danfma:
**Priority:** Future

**Background:** [ADR-0001](../blob/main/docs/adr/0001-target-agnostic-core.md) establishes that each language target is its own project referencing \`Metano.Compiler\` (the target-agnostic core). Adding a new target should be additive: new project, new AST, new handlers, no changes to the core or to sibling targets.

## Goal

A new \`Metano.Compiler.Dart\` project that transpiles annotated C# types into idiomatic Dart, targeted at Flutter consumers. Same domain types (records, enums, interfaces, LINQ-style queries) shared between a .NET backend and a Flutter frontend without hand-maintained mirror code.

## Scope (very high level — this is a big project)

- [ ] New project \`src/Metano.Compiler.Dart/\` implementing \`ITranspilerTarget\`
- [ ] Dart-specific AST (\`DartType\`, \`DartClass\`, \`DartMethod\`, etc.) + Printer
- [ ] CLI tool: \`metano-dart\`
- [ ] Lowering rules for the core constructs:
    - C# records → Dart \`class\` with \`==\` / \`hashCode\` / \`copyWith\`
    - C# enums → Dart \`enum\` (already native)
    - Nullable types → Dart \`T?\`
    - Generics with constraints
    - Async / await (\`Task<T>\` → \`Future<T>\`)
- [ ] BCL mappings for Dart: \`DateTime\` → \`DateTime\`, \`decimal\` → package (\`decimal\` on pub.dev), \`Guid\` → \`String\` or a branded wrapper
- [ ] LINQ mappings (Dart has \`Iterable\` with some LINQ-flavoured methods, but not all — some will need runtime helpers)
- [ ] \`[EmitPackage]\` support for Dart pub packages
- [ ] Sample project \`samples/SampleTodo.Dart/\` with generated output and a Flutter consumer
- [ ] CI integration — Dart SDK in the workflow, \`flutter test\` on the generated Flutter sample

## Notes

This is a multi-month project. The goal of the issue is to have a single place to track scope, design discussion, and sub-task breakdown when someone decides to pick it up.

Part of #14.

danfma:
**Spec mapping:** Spec change request — new target. Would need a new section in the spec for Dart-specific requirements. The target-agnostic core (FR-035, FR-037) already supports this architecturally (ADR-0001).

PR #168, opened 2026-04-29
feat(dart): land Dart target foundation — runtime, imports, BCL, delegates

danfma:
## Summary

Lands the foundational layers of the Dart/Flutter target on top of the prototype merged earlier (`f008268`). Six commits, each a self-contained phase mirroring the corresponding TypeScript target capability so the two backends share extraction, IR shapes, and the declarative mapping pipeline.

## Phases (commit-by-commit)

1. **`feat(dart): populate metano_runtime with hashing and equality primitives`** — Port of xxHash32 from the TS runtime (bit-identical, golden tests cross-check), `deepEquals` / `deepHashCode` for collections, expanded `console` (warn/error via `dart:developer`). Foundation for every consumer that follows.

2. **`feat(dart): add structured DartImportCollector wired to runtime requirements`** — Extracts the inline import walker out of `DartTransformer` into a dedicated `DartImportCollector` and feeds it `IrRuntimeRequirementScanner` output, so generated Dart files declare a `package:metano_runtime` import whenever the IR signals it. Mirrors the TS backend's collector + `IrRuntimeRequirementToTsImport` split.

3. **`feat(dart): inject MetanoObject base and route hashCode through runtime HashCode`** — Records synthesize hashCode through the runtime helper (`HashCode.combine[N]` for up to four fields, builder API beyond), replacing the broken `Object.hash` codegen that emitted invalid single-arg calls for one-field records.

4. **`refactor(dart): drop MetanoObject auto-injection and runtime base`** — Removes the implicit Metano-only base class added in (3). Generated records now sit directly under Dart's `Object`, which keeps interop with native Dart types and external packages clean. The HashCode codegen from (3) stays — that one earned its place.

5. **`feat(dart): wire declarative BCL mappings via target-aware MapMethod attrs`** — Adds `DartMethod` / `DartTemplate` / `DartRuntimeImports` siblings to `[MapMethod]` / `[MapProperty]` and threads them through the IR, frontend extractor, and a new `DartIrRewriter` walker. The shared `DeclarativeMappingRegistry` moves into `Metano.Compiler`. First end-to-end mapping: `Console.WriteLine` → `print` (rename form). `DartTemplate` is parsed but not yet rendered.

6. **`feat(dart): lower delegates to Dart typedefs`** — Replaces the prototype's diagnostic-fallback for `TypeKind.Delegate` with real emission. Each transpilable C# delegate becomes a Dart `typedef` aliasing the function signature; `[This]`-attributed parameters degrade to a regular positional receiver. The shared `IrToDartTypeParameterMapper` collapses duplication between class and delegate bridges.

## Verification

- `dotnet build Metano.slnx` — green.
- `dotnet run --project tests/Metano.Tests/` — **874/874** passing (25 of those are new from this branch).
- `dotnet run --project src/Metano.Compiler.Dart/ -- -p samples/SampleCounter/SampleCounter.csproj -o targets/flutter/sample_counter/lib/sample_counter --clean` — regenerates four files (Counter, CounterPresenter, ICounterView, Program).
- `cd targets/flutter/sample_counter && flutter analyze` — **No issues found**.
- TypeScript: zero behavioral change. Only structural touch is `DeclarativeMappingRegistry.cs` moving from `Metano.Compiler.TypeScript/Transformation/` to `Metano.Compiler/Transformation/` (rename, content identical) so both targets consume the same factory.

## Sample output

\`\`\`dart
// counter.dart
import 'package:metano_runtime/metano_runtime.dart' show HashCode;

class Counter {
  Counter(this.count);
  final int count;
  static Counter get zero => Counter(0);
  Counter increment() => Counter(this.count + 1);
  Counter decrement() => Counter(this.count - 1);
  @override
  bool operator ==(Object other) => other is Counter && other.runtimeType == this.runtimeType && other.count == this.count;
  @override
  int get hashCode => HashCode.combine(this.count);
  Counter copyWith({int? count}) => Counter(count ?? this.count);
}
\`\`\`

\`\`\`dart
// program.dart
class Program {
  static void main() => print('Hello, World!');
}
\`\`\`

## Out of scope (follow-ups, in priority order)

- `[ModuleEntryPoint]` → top-level Dart `void main()` (currently emits a class wrapper).
- `IrTemplateExpression` rendering in `IrBodyPrinter` so `DartTemplate` mappings actually fire.
- BCL surface expansion (string, List, Dictionary, Math, LINQ extensions) — additive data only, no infra change.
- AST coverage: exception types, `[GenerateGuard]` type guards, `[PlainObject]` shape, overload dispatcher.
- `IrBodyPrinter` (Dart) full coverage so generated bodies stop emitting placeholder comments for unsupported nodes.
- LINQ Enumerable port to `metano_runtime`.
- ADR for `build_runner` integration viability.

Part of #28


danfma:
Addressed the review feedback in `11d7280`:

**P1 — `CSharpSourceFrontend.cs:1872` (codex): Dart-only mappings entering JS registry**
Added `HasJsMapping` to `DeclarativeMappingEntry` and filtered both the call and member-access paths in `IrToTsBclMapper` to skip entries that carry only Dart data. The `JsName!` / `JsTemplate!` dereferences can no longer hit a Dart-only attribute. Symmetric to the existing `HasDartMapping` filter on the Dart side.

**P1 — `Console.cs:15` (codex): `Console.WriteLine` arity mismatch**
Added `WhenArgCount` to `MapMethodAttribute` (and the IR entry / both mappers). The `Console.WriteLine → print` mapping now uses `WhenArgCount = 1`, so format-string overloads stay un-mapped on the Dart side instead of lowering to the invalid `print(format, args...)` shape. New regression test `DartMethodMapping_HonorsWhenArgCountFilter` pins the behavior.

**Doc — `docs/better_flutter_support_plan.md` (gemini, copilot)**
Rewrote the document in English and replaced the original aspirational draft with a current-status snapshot. The reverted `MetanoObject` base-class injection is now captured under "Architectural decisions" with the rationale, and the remaining phases (5b–5f, BCL surface, body printing, runtime gaps, build_runner spike) are tracked in the status table.

**i18n — `targets/dart/metano_runtime/lib/src/collections.dart` (copilot)**
Translated the placeholder comment to English to match the rest of the runtime sources.

Build green, full suite **875/875** passing, sample regen + `flutter analyze` clean.

PR #196, opened 2026-05-06
feat(compiler,ts): Phase B — capture lambda body as IR expression tree for IQueryable (#31)

danfma:
## Summary

Phase B of #31. Walks each LINQ stage's lambda body when the call opts into queryable capture and emits an `IrQueryableMeta` alongside the closure. The TypeScript bridge lowers the meta into a `{ tree, captures? }` object literal as the trailing argument to the pipe operator, so a runtime IQueryable provider can read the body shape without opening the closure.

**Stacked on #195** — depends on the LINQ pipe runtime + `IrLinqChain` lowering.

## Capture triggers

- Receiver implements `System.Linq.IQueryable<T>`
- Stage method or argument parameter carries `[Queryable]`
- Argument parameter type is `Expression<Func<…>>`

## Walker MVP subset

`param`, `capture`, `literal`, `member`, `call`, `binary`, `unary`, `conditional`. Bodies outside the subset bail out — the lambda flows as a plain closure, no meta emitted. MS0024 reserved in the diagnostic catalog for the future hard-error path.

## Pipeline

- `IrExpressionTreeExtractor` — Roslyn lambda body → `IrExprTreeNode`
- `IrExpressionExtractor.TryCaptureQueryableMeta` — detects trigger, invokes walker on the principal lambda
  - `ShouldCaptureExpressionTree` separates detection from invocation
  - `ResolveStageParameter` handles extension-method `ReducedFrom` offset
- `IrToTsExpressionBridge.MapQueryableMeta` + `MapExprTree` — emits the runtime object literal; `TreeNode(kind, fields...)` helper keeps each variant a single line
- CamelCase applied at the bridge level; IR keeps PascalCase member/method names so other targets (Dart) can apply their own casing

## Example output

`users.Where(u => u.Age >= threshold && u.Active)` with `IQueryable<User>` source:

```ts
linq(users, where((u: User) => u.age >= threshold && u.active, {
  tree: { kind: \"binary\", op: \"&&\",
    left: { kind: \"binary\", op: \">=\",
      left: { kind: \"member\", target: { kind: \"param\", name: \"u\" }, member: \"age\" },
      right: { kind: \"capture\", name: \"threshold\" } },
    right: { kind: \"member\", target: { kind: \"param\", name: \"u\" }, member: \"active\" } },
  captures: { threshold: threshold },
}));
```

## Reviews applied (compiler-man + bob)

- Capture key normalized via `OriginalDefinition` (handles generic constructed symbols)
[...]

Issue #31, opened 2026-04-11
Lambda and expression tree capture (IQueryable-like)


danfma:
**Priority:** Future

**Background:** Currently, C# lambdas lower directly to TypeScript arrow functions — the AST is lost at emit time, so the generated code can't inspect a lambda's structure at runtime. This forecloses a family of features that depend on representing lambdas as data.

## Goal

Capture lambdas in C# \`Expression<Func<T,U>>\` form (or a Metano-specific equivalent) as reified TS AST objects that a runtime visitor can walk. Unlocks the \`IQueryable\` pattern: take a user-written predicate, translate it into SQL / GraphQL / a JSON filter body, or apply optimizations before executing it.

## Scope

- [ ] Detect \`Expression<Func<...>>\` parameter / property types in the C# source
- [ ] Emit the captured expression as a TS data structure (tagged-union AST) instead of an arrow function
- [ ] Provide a runtime visitor API in \`metano-runtime\` that consumers can subclass to translate expressions into their target language (SQL, GraphQL, HTTP query params, etc.)
- [ ] Build-time optimizations:
    - Inline simple lambdas (constant folding)
    - Fuse adjacent operators (\`.where(p).where(q)\` → \`.where(p && q)\`)
    - Hoist pure sub-expressions
- [ ] A sample that uses the pattern to translate LINQ-style queries into a custom backend (e.g. an in-memory \`IQueryable\`-over-arrays)
- [ ] Security note: if \`string + eval\` ever comes up as an implementation shortcut for dynamic expression construction, **do not ship it**. Reified AST only.

## Notes

This is a research-scope issue. Start by picking one concrete downstream use case (SQL? GraphQL? filter-query DSL?) and designing the visitor API for that case. Generalizing comes after the first working implementation.

Part of #14.

danfma:
**Spec mapping:** Spec change request — expression tree capture is a new feature area not covered by any current FR. Would need FR for reified lambda AST + runtime visitor API.

PR #197, opened 2026-05-07
feat(sample): demo IQueryable provider over arrays via Phase B trees (#31)

danfma:
## Summary

Concrete demonstration of the Phase B (#196) `QueryableMeta` shape: a custom provider walks each LINQ stage's expression tree and materializes the result over an in-memory array, bypassing the runtime closures.

The sample doubles as a feedback signal — building it surfaced two compiler bugs in Phase B's lowering that landed as fixes in this same branch.

## What's in the box

- **`samples/SampleQueryableArrays/`** — C# domain (`User` record + `UserQueries` static class) with four queries: `adults`, `activeAdults`, `adultsAtLeast` (closure capture), `adultNames` (where + select).
- **`targets/js/sample-queryable-arrays/`** — generated TS + hand-written `array-provider.ts` that walks the tree (`param`/`capture`/`literal`/`member`/`call`/`binary`/`unary`/`conditional`) and 5 bun tests cross-checking provider results vs the runtime closures.

The provider returns `{ materialized, fellBack }` so consumers can detect coverage gaps — stages with no `queryable` meta still run (via `apply`) and the kind is logged in `fellBack`.

## Compiler fixes (found via the sample)

- **`IrExpressionExtractor.BuildLinqChain`**: drill through `AsQueryable()` calls. The BCL uses them only to bind the next stage to `System.Linq.Queryable` so lambda params become `Expression<Func<…>>`. The runtime treats both `IEnumerable` and `IQueryable` as plain `Iterable`, so the chain walker now drops `AsQueryable` and keeps its receiver as the chain source. Without this, generated code emitted `source.asQueryable()` which has no runtime counterpart.
- **`IrToTsExpressionBridge.MapOptionalType`**: trim trailing newline from `Printer.RenderType` output (`IndentedStringBuilder.ToString` appends one). Otherwise queryable tree nodes emitted `type` fields like `\"User\\n\"` which broke biome formatting on consumer side.

## Test plan

- [x] `dotnet run --project tests/Metano.Tests/` — 974 tests pass
- [x] `cd targets/js/sample-queryable-arrays && bun run build && bun test` — 5 provider tests pass
- [x] No regressions in sample-todo (18), sample-issue-tracker (142), sample-todo-service (33)

## Follow-up

Sets the stage for #31 (visitor API) — once we know what a real provider needs, we can extract the common bits of `array-provider.ts` (tree walking, scope resolution, capture lookup) into a reusable subclassable base in `metano-runtime/system/linq`.

Part of #31.

### DIFF
--- a/Metano.slnx
+++ b/Metano.slnx
@@ -23,5 +23,6 @@
     <Project Path="samples\SampleCounterV1\SampleCounterV1.csproj" />
     <Project Path="samples/SampleCounterV3/SampleCounterV3.csproj" />
     <Project Path="samples/SampleCounterV4/SampleCounterV4.csproj" />
+    <Project Path="samples/SampleQueryableArrays/SampleQueryableArrays.csproj" />
   </Folder>
 </Solution>
--- a/samples/SampleQueryableArrays/AssemblyInfo.cs
+++ b/samples/SampleQueryableArrays/AssemblyInfo.cs
@@ -1,0 +1,4 @@
+using Metano.Annotations;
+
+[assembly: TranspileAssembly]
+[assembly: EmitPackage("sample-queryable-arrays", Version = "workspace:*")]
--- a/samples/SampleQueryableArrays/SampleQueryableArrays.csproj
+++ b/samples/SampleQueryableArrays/SampleQueryableArrays.csproj
@@ -1,0 +1,17 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFramework>net10.0</TargetFramework>
+    <IsPackable>false</IsPackable>
+    <MetanoOutputDir>../../targets/js/sample-queryable-arrays/src</MetanoOutputDir>
+    <MetanoClean>true</MetanoClean>
+    <MetanoFilePrefix>/** biome-ignore-all lint/complexity/noUselessConstructor: explicit shape preserved by transpiler */</MetanoFilePrefix>
+    <MetanoCompilerProject>../../src/Metano.Compiler.TypeScript/Metano.Compiler.TypeScript.csproj</MetanoCompilerProject>
+    <MetanoPostBuildCommand>bunx biome check --write &quot;$(MetanoOutputDir)&quot;</MetanoPostBuildCommand>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="../../src/Metano/Metano.csproj" PrivateAssets="all" />
+  </ItemGroup>
+
+  <Import Project="../../src/Metano.Build/build/Metano.Build.targets" />
+</Project>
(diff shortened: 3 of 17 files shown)


danfma:
@gemini-code-assist @chatgpt-codex-connector @copilot-pull-request-reviewer — addressed in 168c136:

- **AKsVj / AKtlZ** (RenderType trailing newline): root-caused at the source. `IndentedStringBuilder` now exposes `GetRawString()` (verbatim buffer, no trailing-newline normalization). `Printer.PrintTypeOnce` / `PrintExpressionFragment` / `PrintType` use it, so `Printer.RenderType` no longer adds a `\n` (or `\r\n` on Windows). The `TrimEnd('\n')` workaround in `MapOptionalType` is gone.
- **AKsVm** (`as never` non-idiomatic): replaced. Comparison ops (`<`, `<=`, `>`, `>=`) route through a `compareValues(left, right): number` helper that casts to `number | string | bigint`. Arithmetic ops cast to `number` explicitly. No more `as any` / `as never` in evaluateBinary.
- **AKsVt** (int division parity): documented as a known limitation. C# `int / int` is integer division; JS `/` is float. Bridging needs per-node int-vs-double type info the IR does not currently distinguish (both lower to TS `number`). Provider mirrors JS semantics for now, matching what the closure path emits.
- **AKtlW / AKuSY** (`readParamName` incomplete): rewrote as a complete recursive scan over every node kind (`whenTrue` / `whenFalse`, `call.args` via `firstParamName` helper, `member.target`). Returns `null` when no param exists; caller defaults to `"x"` so a literal-only body still materializes. The earlier shape silently dropped param references in conditional branches and call arguments.
- **AKuR_** (AsQueryable static-call form): `ResolveAsQueryableInner` now covers both shapes — extension (`items.AsQueryable()`, receiver is the source) and static (`Queryable.AsQueryable(items)`, first argument is the source). Static form is detected when the member-access receiver resolves to an `INamedTypeSymbol`; falls back to `current.ArgumentList.Arguments[0].Expression`.
- **AKuSh** (`evaluateMember` silent undefined): now throws a `TypeError` when the target is null/undefined, matching plain JS property-access semantics and the closure path the provider cross-checks against.
- **AKuSo** (`ExprCall.target === null`): no longer crashes with a misleading 'no resolved target' message. Surfaces as 'free-function call is not supported in this MVP provider' so a higher-level orchestrator can decide whether to fall back.
- **AKuSs** (`compilePredicate` misnamed): renamed to `compileLambdaBody` — it compiles both predicates (where) and selectors (select).

Visitor API extraction (gemini AKsVj / sample duplication) tracked separately as #198 — generic walker stays in the sample for now per scope.

Tests: 974 .NET, 5 sample-queryable-arrays, all other sample suites green.

Issue #200, opened 2026-05-07
Expose chain stages on linq() result for IQueryable provider introspection


danfma:
## Background

Phase B (#196) attaches \`QueryableMeta\` (the IR expression tree) to each operator descriptor produced by the runtime's \`where\` / \`select\` / \`orderBy\` / etc factories. The descriptors are passed to \`linq(source, ...stages)\` which executes the chain via \`apply\` reduction and returns the materialized \`Iterable<T>\` (or a scalar, for terminal-bearing chains).

After \`linq()\` returns, the \`stages\` array is gone. Sample providers (\`sample-queryable-arrays\`, the upcoming \`sample-queryable-sqlite\`) work around this by intercepting stages via a \`captureStages\` helper in tests — fine for unit testing, awkward for a real consumer who simply wants to do:

\`\`\`ts
const query = repo.products.where((p) => p.isActive); // returns IQueryable<Product>
const rows = repo.toArray(query);                      // adapter inspects stages, emits SQL
\`\`\`

## Proposal

Attach the descriptor chain to the result of \`linq()\` as a non-enumerable symbol-keyed property so introspecting consumers can read it without the runtime exposing a new top-level entry point or breaking the existing \`Iterable<T>\` shape.

\`\`\`ts
const STAGES = Symbol.for(\"metano-runtime/linq:stages\");

export function linq(source, ...ops): Iterable<unknown> {
  const result = ops.reduce((acc, op) => op.apply(acc), source);
  if (typeof result === \"object\" && result !== null) {
    Object.defineProperty(result, STAGES, { value: ops, enumerable: false });
  }
  return result;
}

export function getStages(query: unknown): Stage[] | undefined {
  return query && typeof query === \"object\" ? (query as any)[STAGES] : undefined;
}
\`\`\`

## Scope

- [ ] Runtime change in \`metano-runtime/system/linq/linq.ts\`.
- [ ] Public \`getStages\` helper exported from the same module.
- [ ] Migrate \`sample-queryable-arrays\` provider to use \`getStages\` instead of \`captureStages\`.
- [ ] Document the contract on the runtime README + add a runtime test that the symbol survives through every operator factory.

## Non-goals

- Changing the public \`linq()\` signature (still returns \`Iterable<T>\` / scalar).
- Replacing the closure path — stages remain a side channel for consumers that opt in.

## Motivation

[...]

Issue #199, opened 2026-05-07
Copy assembly assets / resources to the generated package output


danfma:
## Background

C# projects can carry non-code assets (SQL schema files, JSON fixtures, static resources, README snippets) that the runtime needs at execution time. Today the Metano transpiler emits TS sources only — anything declared as \`<EmbeddedResource>\` / \`<Content>\` / \`<None CopyToOutputDirectory=\"…\" />\` in the csproj never reaches the generated package directory.

Surfaced while building \`samples/SampleQueryableSqlite\` (#197 follow-up) — the SQLite schema DDL would naturally live as a \`schema.sql\` resource alongside \`Product.cs\`, but there is no way to ship it into \`targets/js/sample-queryable-sqlite/src/\` automatically. The sample worked around it by inlining the DDL as a C# string constant; that's fine for a small fixture but does not generalize.

## Scope

- [ ] Define a csproj-level convention for declaring \"copy this asset to the package output\". Most natural fit is a new MSBuild item type (\`<MetanoAsset Include=\"…\" />\`) or honoring \`<None CopyToOutputDirectory>\`.
- [ ] Resolve the destination relative to \`MetanoOutputDir\` (mirroring the source folder structure).
- [ ] Skip clean (\`MetanoClean=true\`) for assets that pre-exist on disk to avoid wiping user-authored files. Or restrict cleaning to known generated extensions (\`.ts\`, \`.d.ts\`).
- [ ] Document the convention in the README + add a sample exercising it (e.g. \`SampleQueryableSqlite\` shipping \`schema.sql\` once this lands).
- [ ] Cross-target story: applies equally to Dart target (\`Metano.Compiler.Dart\`).

## Non-goals

- Bundling assets into the TS module graph (that's the consumer's bundler concern).
- Asset transformation pipelines (sass, image optimization, …).

## Workaround until this lands

Inline the asset as a C# string literal / static field in a transpilable type. Acceptable for small fixtures; lossy for binary or large content.

## References

- \`src/Metano.Build/build/Metano.Build.targets\` — current build pipeline.
- Sample that motivated this: \`samples/SampleQueryableSqlite\` (PR pending).

Issue #198, opened 2026-05-07
Generic ExprTree visitor API in metano-runtime for IQueryable provider authors


danfma:
## Background

Phase B (#196) captures every LINQ stage's lambda body as an `ExprTree` IR alongside the closure. PR #197 demonstrates how a consumer can read that tree to materialize a query without opening the closure — the `sample-queryable-arrays` sample contains a hand-rolled walker (`evaluate`, `compileLambdaBody`, `readParamName`, `compareValues`) that any other provider would have to duplicate.

## Problem

The walker has two layers:

- **Generic**: recursive dispatch over the `ExprTree` union (param/capture/literal/member/call/binary/unary/conditional), scope/captures resolution, primitive comparator, lambda-body compilation. Reusable across every JS-side provider.
- **Specific**: how the materialized predicate/selector applies to the backing store (filter an array, build a SQL `WHERE` clause, push HTTP query params, …).

Real providers (a SQL translator, a GraphQL builder, an in-memory mock) would all reimplement the generic layer today.

## Proposal

Publish a stable visitor API from `metano-runtime/system/linq` so providers extend or compose it instead of forking the sample:

- **`evaluateExprTree(tree, scope)`** — direct JS evaluation; default behavior used by mock providers.
- **`compileLambdaBody(meta, fallbackParamName?)`** — closes over captures, returns `(item) => unknown`.
- **`ExprTreeVisitor<R>`** (subclassable) — abstract base with one `visitX` method per node kind, default implementations that recurse, so subclasses override only what they care about (e.g. SQL translator overrides `visitBinary` to emit operators).
- **Helpers**: `readParamName`, `compareValues`, scope plumbing.

## Scope

- [ ] Design the visitor API (subclass vs functional vs both — driven by the SQL/GraphQL use cases that motivated #31).
- [ ] Move the sample's walker into the runtime; reduce `sample-queryable-arrays/src/provider/array-provider.ts` to ~50 lines (just `handleWhere`/`handleSelect` + `runArrayProvider` orchestrator).
- [ ] Cover the runtime walker with `metano-runtime/test/` cases (param resolution, capture lookup, conditional, nested member, primitive comparator, fallback param name).
- [ ] Document the surface in the runtime README + add a short note in `CLAUDE.md` linking the queryable attribute + the visitor API.

## Non-goals (this issue)

- Building an actual SQL/GraphQL provider — that lives in a separate sample once the visitor base lands.
[...]

PR #201, opened 2026-05-07
feat(sample,runtime,compiler): SQLite EF-lite provider via captured expression trees

danfma:
## Summary

Adds `samples/SampleQueryableSqlite` + `targets/js/sample-queryable-sqlite`, a thin EF-lite-style ORM that translates `IQueryable<Product>` chains into a single parameterized SQL statement and executes them via Bun's built-in `bun:sqlite` driver.

The C# side declares the domain (\`Product\`), the adapter contract (\`IProductRepository\`), and a demo class (\`ProductDemo\`) that composes LINQ chains over \`IQueryable<Product>\` and hands them to the adapter. The TypeScript side implements the adapter, owning schema bootstrap, snake_case column mapping, the SQL translator, and the actual `bun:sqlite` execution.

This sample exercises a lot of Metano's interop surface in one place — \`[Transpile]\` interfaces, \`[Import]\` for Bun runtime modules, \`[External]\` + \`[Emit]\` for variadic JS APIs, IQueryable-driven Phase B capture, and the new runtime stage-introspection helper.

### C#

- \`Product.cs\` — record entity (id, name, displayName, unitPrice, stockCount, isActive).
- \`Provider/IProductRepository.cs\` — \`[Transpile]\` interface declaring \`Products: IQueryable<Product>\` + \`ToArray\` / \`Count\` / \`FirstOrDefault\` materializers.
- \`Provider/ProductDemo.cs\` — \`[Transpile]\` static class with example queries (\`ActiveProducts\`, \`InStock\`, \`AtLeastPrice\`, \`TopByPrice\`, \`Page\`, \`CountActive\`, \`FirstById\`).
- \`Js/BunSqlite/Database.cs\` — \`[Import(name: \"Database\", from: \"bun:sqlite\")]\` façade.
- \`Js/BunSqlite/Statement.cs\` — \`[External]\` interface + \`[Emit]\` extensions that lower variadic \`params object?[]\` arguments to JS rest spread (\`stmt.all(...args)\`).

### TypeScript

- \`src/provider/db.ts\` — schema DDL, seed, snake_case column map, \`mapProductRow\`.
- \`src/provider/sqlite-translator.ts\` — pure ExprTree → SQL walker covering \`where\` / \`select\` (single member projection) / \`orderBy[Descending]\` / \`thenBy[Descending]\` / \`take\` / \`skip\` + \`count\` / \`first\` / \`firstOrDefault\` terminals.
- \`src/provider/bun-sqlite-product-repository.ts\` — concrete \`IProductRepository\` that reads stages via \`getStages\`, translates, and executes; falls back to closure path on \`UntranslatableTreeError\`.
- \`test/product-demo.test.ts\` — 7 cross-checks against an in-memory golden over the seed.

### Compiler

- \`IrTypeRefMapper\`: lower \`IQueryable<T>\` to \`Iterable<T>\` alongside \`IEnumerable<T>\` / \`IReadOnlyCollection<T>\`. The runtime treats both the same; the C# distinction only triggers Phase B tree capture.

### Runtime (#200)

- \`linq()\` now attaches the ops array on its object-typed result via the \`LINQ_STAGES\` symbol so an introspecting consumer can read the chain without monkey-patching or hand-rolling a capture helper.
- New \`getStages(query)\` helper returns the descriptor list for a query result, undefined for primitives or non-runtime values.
- Two runtime tests cover the symbol slot + scalar-result fallback.

## Tracking issues opened along the way

- **#198** — generic ExprTree visitor API in \`metano-runtime\`. The repository's translator could move there once a couple of real providers exist.
- **#199** — copy assembly assets / resources to the generated package output. Surfaced because the schema DDL would naturally ship as \`schema.sql\` next to \`Product.cs\`; for now the DDL is inlined as a TS template literal.
- **#200** — runtime stage introspection helper. Landed in this PR.

## Test plan

- [x] \`dotnet run --project tests/Metano.Tests/\` — 974 tests pass (no new tests on the .NET side; the sample exercises existing transpile paths).
- [x] \`metano-runtime\`: 246 tests pass (+2 new \`getStages\` cases).
- [x] \`sample-queryable-sqlite\`: 7 tests pass.
- [x] No regressions in \`sample-todo\` (18), \`sample-issue-tracker\` (142), \`sample-queryable-arrays\` (5), \`sample-todo-service\` (33).

## Out of scope

- SQL JOIN / GroupBy / aggregate translators.
- Transaction / connection-pool story.
- Reflection-based mapping. The column map is hand-authored, mirroring how EF Core's \`OnModelCreating\` works but simpler.
- Generic ORM library. The repository is concrete to \`Product\`; generalizing is the natural next step (and aligns with #198).


danfma:
@gemini-code-assist @chatgpt-codex-connector @copilot-pull-request-reviewer — addressed in c3c0c32:

- **AVXKQ / AVZt0** (count + skip/take semantics): the translator now throws `UntranslatableTreeError` when `count()` is combined with `take` or `skip`. The repository catches the throw and falls back to the closure path, so chains like `Skip(5).Take(10).Count()` materialize via the iterator and return the correct windowed count instead of silently dropping the LIMIT/OFFSET.
- **AVXKe / AVZts** (skip without take): emit `LIMIT -1` whenever `OFFSET` is set without an explicit `LIMIT`. SQLite rejects the bare `OFFSET n` form; `-1` is the documented no-upper-bound sentinel.
- **AVcm1** (root package.json contamination): restored the workspace root identity. The Metano package.json writer climbed past the missing sample package.json on the first generation and overwrote the workspace root; the sample's own package.json now exists so future regenerations land in the right place.
- **AVcoX** (projection by `Object.values(row)[0]`): the translator now aliases the projected column to a stable `PROJECTION_ALIAS` (`value`); the repository reads the value by key. The `Product[]` return-type cast is acknowledged in a comment — the API today only exercises projection through chains like `ActiveDisplayNames` where the C# user explicitly opts into the scalar shape.
- **AVco2** (pushOrder error message): the helper now takes the actual stage kind so `UntranslatableTreeError` reports `'thenBy'` / `'orderByDescending'` etc. instead of always saying `'orderBy'`.
- **AVcpT** (IOrderedQueryable / IOrderedEnumerable): `IrTypeRefMapper` now lowers both marker interfaces to `Iterable<T>` alongside `IEnumerable<T>` / `IQueryable<T>`. `OrderBy` + `ThenBy` chains keep their TS surface intact.
- **AVcnp** (Decimal equality in generated equals()): tracked separately as **#202**. The fix is a compiler change spanning every BCL value-wrapper (Decimal, Temporal, etc.) — out of scope for this PR.

Tests: 974 .NET, 7 sample-queryable-sqlite, biome clean.

PR #204, opened 2026-05-07
fix(compiler,runtime): record equals() routes value-wrapper fields through valueEquals

### LINKED ISSUE #202: Generated record equals() compares value-wrapper fields by reference (Decimal, Temporal, etc.)
## Background

C# \`record\` types lower to TS classes whose \`equals(other)\` method compares every field with strict equality (\`this.x === other.x\`). For primitive fields this matches C# semantics; for value-wrapper types like \`decimal.js\` \`Decimal\` (the BCL mapping for C# \`decimal\`) and Temporal types it does not — two numerically equal \`Decimal\` instances are different objects, so \`equals\` returns \`false\` even when the records would compare equal in C#.

Surfaced on \`SampleQueryableSqlite.Product\` (#201) — \`Product.UnitPrice\` is a \`decimal\` field; the generated \`equals\` compares the two \`Decimal\` instances by reference and silently diverges from .NET semantics. The same wrapper hits Temporal types (\`PlainDate\`, \`PlainDateTime\`, \`Duration\`) once those flow through a record.

## Proposal

For each record field, the bridge should pick the comparison strategy from the field's lowered type:

- **Primitive (number / string / bigint / boolean):** \`===\`.
- **Value-wrapper with a known \`equals\`-shaped method (Decimal, Temporal, HashSet/HashMap, our own \`[Branded]\`/\`[InlineWrapper]\` types):** call \`x.equals(y)\` (or the type-specific helper, e.g. \`Decimal.eq(a, b)\`, \`Temporal.PlainDate.compare(a, b) === 0\`).
- **Reference type:** keep \`===\` (current behavior — matches reference-equality for nullable record fields).

The bridge already inspects field types at lowering time (\`IrToTsExpressionBridge\` / record handler); the comparison choice can fold into the same dispatch.

## Scope

- [ ] Identify the comparison helper per BCL mapping (Decimal: \`.eq()\`, Temporal: \`.equals()\` / \`.compare() === 0\`, HashSet / HashMap: structural).
- [ ] Update the record \`equals\` lowering to pick the right helper per field type.
- [ ] Update \`hashCode\` accordingly so the contract stays consistent — use the wrapper's stable hash instead of the wrapper's identity.
- [ ] Tests for Decimal / Temporal record fields.

## Workaround until this lands

Override \`equals\` / \`GetHashCode\` manually on records carrying value-wrapper fields. For DTOs that flow through JSON, the issue is moot because the wire shape is plain — only in-memory comparisons hit the bug.

## References

- #196 / #197 — captured trees + array-provider sample.
- #201 — SQLite sample where \`Product.UnitPrice\` exposes the bug.

danfma:
## Summary

Fixes #202. C# record positional parameters whose lowered TS type is a JS object-wrapper (\`Decimal\` from \`decimal.js\`, Temporal date / time, the runtime \`UUID\`, nested transpiled records, plain classes) used to compare with strict \`===\` inside the synthesized \`equals\` method. Two numerically equal \`Decimal\` instances are different objects, so the record's equals contract silently diverged from C# value-equality semantics.

The compiler now picks a per-field comparison strategy based on the IR type and emits \`valueEquals(this.x, other.x)\` for wrapper fields, paired with a runtime import line.

## Strategy

| IR type                                            | Comparison                  |
|----------------------------------------------------|-----------------------------|
| \`Boolean\`, \`Byte\`, \`Int16/32/64\`, \`Float32/64\`, \`BigInteger\`, \`String\`, \`Char\`, \`Object\` | \`===\` (real JS primitives) |
| \`Decimal\`, \`Guid\`, \`DateTime\`, \`DateTimeOffset\`, \`DateOnly\`, \`TimeOnly\`, \`TimeSpan\` | \`valueEquals\` (JS object wrappers) |
| \`StringEnum\`, \`NumericEnum\`, \`Branded\`         | \`===\`                      |
| Nested transpiled records / classes / interfaces   | \`valueEquals\`              |
| Type parameters (\`T\`, \`TKey\`)                    | \`===\` (preserve baseline)  |
| Arrays, maps, sets, iterables, tuples, …           | \`valueEquals\`              |

\`valueEquals\` short-circuits on \`a === b\` before delegating to the wrapper's \`.equals(other)\` method, so the wrapper-aware path stays correct for primitives that happen to flow through a generic field.

### Compiler

- **\`IrToTsRecordSynthesisBridge\`**: per-field equality dispatch. Looks up the IR type for each ctor parameter by positional index (TS names are camelCased while the IR keeps the C# casing) and routes wrapper fields through \`valueEquals\`.
- **\`IrRuntimeRequirementScanner\`**: mirrors the bridge's strictness whitelist so a record carrying any non-strict field emits \`IrRuntimeRequirement(\"ValueEquals\")\`.
- **\`IrRuntimeRequirementToTsImport\`**: maps the new requirement to \`import { valueEquals } from \"metano-runtime\"\`.

### Runtime

- **\`metano-runtime/system/equality.ts\`**: re-exports the existing \`equals\` helper as \`valueEquals\` so generated call sites read unambiguously next to the surrounding \`equals\` method declaration. Documented the new compiler contract.

### Samples

- \`sample-queryable-sqlite/src/product.ts\` regenerates with \`valueEquals(this.unitPrice, other.unitPrice)\` (Decimal field). All other samples regenerate unchanged because none had a wrapper field on the synthesized records.

## Test plan

- [x] **978 .NET tests pass** (was 974, +4 new \`RecordValueEqualityTests\` covering Decimal, DateTime, nested record, primitive-only baseline). \`EndToEndOutputTests\` golden refreshed for the Decimal field.
- [x] **246 runtime tests pass** (no behavior change — \`valueEquals\` is just a re-export).
- [x] No regressions in \`sample-todo\` (18), \`sample-issue-tracker\` (142), \`sample-queryable-arrays\` (5), \`sample-queryable-sqlite\` (7), \`sample-todo-service\` (33).

## Out of scope

- \`hashCode\` synthesizing a wrapper-aware hash. The current \`HashCode.add\` helper relies on the field's own \`hashCode()\` contract for objects, which Decimal / Temporal already implement, so this PR keeps the hashing path untouched. A follow-up could enforce the same dispatch for symmetry.
- Cross-target story. Dart / Kotlin targets have their own equality conventions; this PR only updates the TypeScript bridge.

### DIFF
--- a/.husky/task-runner.json
+++ b/.husky/task-runner.json
@@ -24,6 +24,12 @@
       "group": "pre-push",
       "command": "bash",
       "args": ["-c", "bunx biome check ."]
+    },
+    {
+      "name": "verify-targets-clean",
+      "group": "pre-push",
+      "command": "bash",
+      "args": ["scripts/verify-targets-clean.sh"]
     }
   ]
 }
--- a/scripts/verify-targets-clean.sh
+++ b/scripts/verify-targets-clean.sh
@@ -1,0 +1,21 @@
+#!/usr/bin/env bash
+# Pre-push gate that mirrors the CI 'Verify samples are regenerated cleanly'
+# step — runs `dotnet build` for the whole solution and fails if any file
+# under targets/ ends up dirty (meaning the C# source was changed without
+# committing the regenerated TS output).
+set -euo pipefail
+
+repo_root="$(git rev-parse --show-toplevel)"
+cd "$repo_root"
+
+if ! dotnet build Metano.slnx --nologo -v q; then
+  echo "verify-targets-clean: dotnet build failed — fix the build before pushing." >&2
+  exit 1
+fi
+
+if ! git diff --quiet -- targets/; then
+  echo "verify-targets-clean: generated TS under targets/ is out of sync." >&2
+  echo "Run 'dotnet build Metano.slnx', commit the regen, and push again." >&2
+  git status --short -- targets/ >&2
+  exit 1
+fi
--- a/src/Metano.Compiler.TypeScript/Bridge/IrRuntimeRequirementToTsImport.cs
+++ b/src/Metano.Compiler.TypeScript/Bridge/IrRuntimeRequirementToTsImport.cs
@@ -58,6 +58,7 @@ public static class IrRuntimeRequirementToTsImport
         req.HelperName switch
         {
             "HashCode" => ("metano-runtime", "HashCode", false),
+            "ValueEquals" => ("metano-runtime", "valueEquals", false),
             "Enumerable" => ("metano-runtime", "Enumerable", false),
             "HashSet" => ("metano-runtime", "HashSet", false),
             "UUID" => ("metano-runtime", "UUID", false),
--- a/src/Metano.Compiler.TypeScript/Bridge/IrToTsRecordSynthesisBridge.cs
+++ b/src/Metano.Compiler.TypeScript/Bridge/IrToTsRecordSynthesisBridge.cs
@@ -56,16 +56,18 @@ public static class IrToTsRecordSynthesisBridge
             "instanceof",
             new TsIdentifier(GetTypeName(ir))
         );
-        foreach (var param in ctorParams)
+        // The IR ctor param list mirrors the TS-lowered ctorParams in
+        // declaration order, so we lift each TS param's IR type by
+        // positional index rather than by name (TS names are
+        // camelCased while the IR keeps the C# casing).
+        var irTypes = ir.Constructor?.Parameters.Select(p => p.Parameter.Type).ToList();
+        for (var i = 0; i < ctorParams.Count; i++)
         {
+            var irType = irTypes is not null && i < irTypes.Count ? irTypes[i] : null;
             condition = new TsBinaryExpression(
                 condition,
                 "&&",
-                new TsBinaryExpression(
-                    new TsPropertyAccess(new TsIdentifier("this"), param.Name),
-                    "===",
-                    new TsPropertyAccess(new TsIdentifier("other"), param.Name)
-                )
+                FieldEquality(ctorParams[i], irType)
             );
         }
         return new TsMethodMember(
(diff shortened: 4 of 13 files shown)


danfma:
@gemini-code-assist @chatgpt-codex-connector @copilot-pull-request-reviewer — addressed in the latest force-push:

- **AW3nv** (gemini, intermediate list): bridge now indexes `ir.Constructor?.Parameters` directly — no `Select.ToList` per record.
- **AW3n3** (gemini, duplicated logic): extracted `Metano.Compiler.IR.IrEqualityClassifier` as the single source of truth. Both the bridge and `IrRuntimeRequirementScanner` consume it now so the import line and the call sites cannot drift.
- **AW3n8 / AW3oI** (gemini, doc syntax): `equality.ts` switched to TypeDoc `{@link …}` for the cross-symbol references.
- **AW5a8 / AW9o1** (codex P1 + copilot, reference-typed collections): the classifier was too eager. The "non-strict" path is now narrowed — arrays, maps, sets, iterables, generators, promises, functions, `KeyValuePair`, and `Grouping` keep `===`. That preserves C# `EqualityComparer<T>.Default` semantics for those shapes (reference equality unless the user overrides it). Tuples stay non-strict because element-wise comparison matches the C# contract; transpiled records / classes / structs / interfaces stay non-strict so user types' own `equals` contract wins.
- **AW9pf / AW9qZ** (copilot, `IrPrimitive.Object`): `object`-typed record fields now route through `valueEquals`, mirroring `object.Equals` virtual dispatch on the .NET side. Both the bridge and the scanner cover this through the shared classifier.
- **AW9qF** (copilot, scanner / bridge in lockstep): both call into `IrEqualityClassifier.UseStrictEquality` so the requirement / import / emission stay aligned.

Bonus: pre-push hook now also runs a new `verify-targets-clean` task that mirrors the CI 'Verify samples are regenerated cleanly' step (`dotnet build Metano.slnx` + `git diff --quiet -- targets/`). An out-of-sync regen now blocks the push locally instead of waiting for a CI red.

Tests: 978 .NET pass (no shape change to `RecordValueEqualityTests`); 246 runtime pass; sample-issue-tracker / sample-queryable-sqlite regenerated and committed; biome + csharpier clean.

Issue #21, opened 2026-04-11
Incremental compilation and parallel TypeTransformer


danfma:
Two perf improvements to the compiler pipeline that become relevant as project size grows. Grouped in one issue because they share the dependency-graph infrastructure.

## Incremental compilation

- [ ] Compute a content hash per \`.cs\` file (file bytes + Roslyn compilation reference metadata that affects it)
- [ ] Persist the hash → generated-files map to disk between runs (\`.metano/cache.json\` or similar)
- [ ] On subsequent runs, skip types whose inputs (source file, referenced types, mapping metadata) haven't changed
- [ ] Invalidate transitively: if a referenced type regenerates, every type that imports it regenerates too
- [ ] Integration with \`--clean\` (opt-out of cache)
- [ ] Tests covering hit, miss, and invalidation paths

## Parallel TypeTransformer

- [ ] Transpilation is embarrassingly parallel per top-level type — \`TypeTransformer\` can process each type on a separate worker
- [ ] Use \`Parallel.ForEach\` or \`Dataflow\` depending on fan-out characteristics
- [ ] Shared state (symbol maps, \`_transpilableTypeMap\`, \`_crossAssemblyTypeMap\`) needs to be read-only during the parallel phase — audit for mutation
- [ ] \`TypeScriptTransformContext\` already exists as immutable shared state ([ADR-0002](../blob/main/docs/adr/0002-handler-decomposition.md)) — verify it's safe to share across threads
- [ ] Diagnostics collection needs thread-safe aggregation (\`ConcurrentBag<MetanoDiagnostic>\` or similar)

## Combined benefit

Incremental + parallel together make a large monorepo build \`O(changed)\` instead of \`O(total)\`, and the baseline of all-types-changed gets faster by a factor close to the number of cores.

## Dependencies

Shares machinery with #18 (watch mode) — both need to know \"which files would change if this C# file is touched\". Keep the dependency graph layer in the core so both can reuse it.

Part of #14.

danfma:
**Spec mapping:** Maps to NFR-016 (performance) + NFR-017 (compilation speed). Incremental compilation and parallelism are performance improvements to the existing pipeline.

Issue #18, opened 2026-04-11
Watch mode (`--watch`) with FileSystemWatcher


danfma:
Add a \`--watch\` flag to \`metano-typescript\` so the transpiler keeps running and re-emits generated TypeScript whenever a source C# file changes. Useful for tight edit-test loops on the frontend while iterating on domain types in C#.

## Scope

- [ ] \`--watch\` CLI flag on \`metano-typescript\`
- [ ] \`FileSystemWatcher\` on the C# project's source tree (project root + \`<Compile Include>\` globs)
- [ ] Debounce rapid bursts (save hooks, formatter runs) — typical editor save produces multiple events
- [ ] Incremental recompilation: only re-transform the touched files and their dependents, not the whole project
- [ ] Graceful shutdown on SIGINT — don't leave stale files behind
- [ ] Integration with \`bun --watch\` / Vite dev server: ensure the generated \`.ts\` files trigger the downstream watcher cleanly (no write-then-rewrite races)
- [ ] Tests covering the debounce + incremental paths

## Dependencies

Shares machinery with #TBD (incremental compilation) — both need the dependency graph to know what to re-emit when a file changes. Coordinate implementation to avoid duplicating the graph.

Part of #14.

danfma:
**Spec mapping:** Spec change request — no FR covers watch mode. Propose adding FR-047 (watch mode with incremental recompile) to `spec/04-functional-requirements.md`.

PR #211, opened 2026-05-07
feat(compiler): type-level dependency graph backbone for incremental + watch (PR 1/4)

danfma:
## Summary

First slice of the **incremental compilation (#21) + watch mode (#18)** roadmap. Adds the shared dependency-graph infrastructure both features need — without consuming it yet. Subsequent PRs land the parallel transformer (#21 part), the cache + invalidation (#21 part), and the file watcher (#18).

## What this PR adds

- **`Metano.Compiler.DependencyGraph.IrTypeDependencyGraph`** — single graph, type-level, derived from the Roslyn symbols in `IrCompilation.TranspilableTypeEntries`. Both forward (\`DependenciesOf\`) and reverse (\`DependentsOf\` + \`TransitiveDependentsOf\`) edges are precomputed.
- **ADR-0018** — explains the granularity choice (per-type, not per-file), Roslyn-direct discovery, foreign-assembly trade-off. ADR README catches up to ADRs 13–17 that were pinned but never indexed.
- **6 new tests** — direct edges, generic argument unwrapping, transitive closure, self-reference skip, BCL drop, unknown FQN.

## Design

- **Granularity is the type, not the file.** Partials + \`[EmitInFile]\` make file-level granularity lossy.
- **Edges via Roslyn symbols.** \`IrCompilation.Modules\` is currently empty (eager IR-tree population is future work); walking Roslyn symbols directly avoids that dependency and gives the graph builder its own clean entry point.
- **Only edges within the current compilation.** Foreign / BCL types are dropped — they invalidate consumers via a separate metadata-hash key in #21's cache.
- **Both directions precomputed** so the dirty propagation in the cache (#21) is a single closure walk over reverse edges.

## Test plan

- [x] 984 .NET tests pass (+6 new \`IrTypeDependencyGraphTests\`)
- [x] biome + csharpier clean
- [x] No regressions in any sample (verified by the new \`verify-targets-clean\` pre-push gate)

## Roadmap (next PRs in this thread)

1. ✅ **PR 1 — type-level dependency graph** (this PR)
2. **PR 2 — parallel TypeTransformer** (#21 part). Uses the shared state contract that ADR-0002 pins; audits the 39 handlers for mutation; switches the main loop to \`Parallel.ForEach\` with \`ConcurrentBag<MetanoDiagnostic>\`.
3. **PR 3 — incremental cache** (#21 part). Adds \`.metano/cache.json\` keyed by per-type content hash + dependency snapshot; uses \`IrTypeDependencyGraph.TransitiveDependentsOf\` to mark dirty.
4. **PR 4 — watch mode** (#18). \`--watch\` CLI + \`FileSystemWatcher\` + debounce; reuses the cache / graph in memory between events.

Each PR is independent and lands separately.

## References

- ADR-0018 — type-level dependency graph
- Issue #21 — incremental compilation + parallel TypeTransformer
- Issue #18 — watch mode
- ADR-0013 — shared IR as canonical semantic representation

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,7 +1,7 @@
 <claude-mem-context>
 # Memory Context
 
-# [Metano] recent context, 2026-05-07 2:14pm GMT-3
+# [Metano] recent context, 2026-05-07 2:30pm GMT-3
 
 No previous sessions found.
 </claude-mem-context>
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,7 +1,7 @@
 <claude-mem-context>
 # Memory Context
 
-# [Metano] recent context, 2026-05-01 4:49pm GMT-3
+# [Metano] recent context, 2026-05-07 2:14pm GMT-3
 
 No previous sessions found.
 </claude-mem-context>
--- a/docs/adr/0018-type-level-dependency-graph.md
+++ b/docs/adr/0018-type-level-dependency-graph.md
@@ -1,0 +1,89 @@
+# ADR-0018 — Type-level dependency graph as the backbone for incremental + watch
+
+**Status:** Accepted
+**Date:** 2026-05-07
+
+## Context
+
+Two pipeline-perf items in the backlog (#21 incremental compilation, #18 watch
+mode) both need the same answer: "if type T was touched, which types must
+regenerate?". The current pipeline lacks any persistent dependency information —
+every run reprocesses every transpilable type, which is fine for a sample (~30
+types) but scales linearly with project size.
+
+Both features cannot ship without first agreeing on the granularity of "what
+depends on what" and where that information lives. Building a separate graph
+inside the cache subsystem and another inside the watcher would lead to drift,
+slightly different invalidation semantics, and double the maintenance.
+
+## Decision
+
+Introduce `Metano.Compiler.DependencyGraph.IrTypeDependencyGraph` — a single,
+type-level dependency graph derived from the Roslyn symbols carried by
+`IrCompilation.TranspilableTypeEntries`.
+
+- **Granularity is the type, not the file.** A `.cs` file can declare N types;
+  a single type can span N files via `partial`; `[EmitInFile]` collapses N
+  types into a single output file. Tracking dependencies per file would lose
+  precision (over-invalidation) and per-output-file would lose the connection
+  to the C# source. The type FQN — `namespace.Name` from
+  `INamedTypeSymbol.OriginalDefinition` — is the stable key.
+- **Edges come from Roslyn symbols, not from the IR.** `IrCompilation.Modules`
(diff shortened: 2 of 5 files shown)


danfma:
@gemini-code-assist @chatgpt-codex-connector — addressed in 59a1278:

- **AZDTb** (gemini, EmptySet mutability): switched to `ImmutableHashSet<string>.Empty` so a downcast cannot mutate the singleton.
- **AZDT2** + **AZDzg** (gemini + codex P1, FQN format): `QualifiedName` now uses Roslyn's `ToDisplayString` with a `SymbolDisplayFormat` that keeps containing types AND type-parameter names. `Foo` and `Foo<T>` stay distinct keys (no more silent overwrite in `outEdges`); nested types like `Outer.Inner` round-trip correctly instead of collapsing to `Demo.Inner`.
- **AZDTo** (gemini, nested type lookup): `AddIfTranspilable` walks the `ContainingType` chain to roll the reference up to its top-level container before the `ownTypes` lookup. Matches how the cache emits — the top-level entry's output file regenerates as a whole, including every nested type co-located with it. Without this rollup the edge would silently drop.
- **AZDTs** + **AZDzj** (gemini + codex P2, transitive closure on cycles): `TransitiveClosure` now removes the seed at the end. Cycles like `A → B → A` re-add the seed via a neighbor while traversing; the strip keeps the documented 'excludes seed' contract honest regardless of cycle topology.

Tests (3 new, 9 total):
- nested-type reference attributes to top-level container
- same-name + different-arity (`Foo` vs `Foo<T>`) keeps keys distinct
- cyclic graph `A → B → A` excludes seed from transitive set

987 .NET tests pass. Pre-push gate (`verify-targets-clean`) is green.

danfma:
@copilot-pull-request-reviewer — addressed in 59a1278 + ac31f15:

- **AZGw3** (cycle/seed contract): `TransitiveClosure` now strips the seed at the end so `A → B → A` cannot smuggle it back via a neighbor — fixed in 59a1278 alongside the gemini / codex pair.
- **AZGxa** (FQN format + nested rollup): `QualifiedName` switched to a `SymbolDisplayFormat` that keeps containing types AND type-parameter names; `AddIfTranspilable` walks the `ContainingType` chain to attribute nested-type references back to the top-level container the cache emits. Both landed in 59a1278.
- **AZGx5** (signature-only docstring): tightened in ac31f15 — type-level docs now name the exact signature surface (base / interfaces / member types / method return + params + constraints / nested types) and call out that body-level references stay out of this slice per ADR-0018.
- **AZGyP** (cycle test missing): the new `TransitiveDependents_OnCyclicGraph_ExcludesSeed` test in 59a1278 covers exactly that — A↔B cycle, walks from A, asserts B is in the closure and A is not.

987 .NET tests pass, biome + csharpier clean, pre-push gate green.

PR #213, opened 2026-05-07
feat(compiler): parallelize TypeTransformer per file group

danfma:
## Summary

PR 2/4 of the #21 + #18 thread. Switches `TypeTransformer.TransformAll`'s per-file-group loop from `foreach` to `Parallel.For` and makes the three shared mutable sinks thread-safe at their declaration site.

- `_diagnostics`: routed through a private `AddDiagnostic` helper that wraps each `Add` in a `lock`. Sequential pre/post-loop sites untouched.
- `TypeMappingContext.CrossPackageMisses`: now `ICollection<string>` backed by `ConcurrentDictionary<string, byte>.Keys` (set semantics, value slot unused).
- `TypeMappingContext.UsedCrossPackages`: now `IDictionary<string, string>` backed by `ConcurrentDictionary<string, string>`. Single external call site (`BclExportTypeOverrides`) widened to the interface.

Source order preserved through a per-index buffer (`perGroupResults[i]`) — the parallel loop fills any worker order, and a post-loop pass copies non-null entries into `files` in declaration order. AsyncLocal slots on `IrToTsTypeMapper` (`UsingAliases`, `NamedTypeRenames`) are race-free with `Parallel.For` because each worker inherits its own execution context.

ADR-0020 captures the decision; ADR-0002 (handler decomposition) gets a thread-safety addendum so the contract for new bridges is explicit.

### Roadmap (#21 + #18)

- [x] PR 1: type-level dependency graph (#211, merged)
- [x] PR 2: parallel TypeTransformer (this PR)
- [ ] PR 3: incremental cache (uses graph + content hash)
- [ ] PR 4: watch mode (#18 — uses cache + graph + parallel loop)

## Test plan

- [x] `dotnet build Metano.slnx` clean
- [x] 987 .NET tests pass — run 3× back-to-back, no flakes
- [x] sample-todo (18), sample-issue-tracker (142), sample-queryable-arrays (5), sample-queryable-sqlite (7), sample-todo-service (33) all green
- [x] generated TS under `targets/` byte-identical to pre-PR output
- [x] pre-push hook (`csharpier-check`, `biome-check`, `verify-targets-clean`) green

Closes part of #21
Part of #18

### DIFF
--- a/docs/adr/0002-handler-decomposition.md
+++ b/docs/adr/0002-handler-decomposition.md
@@ -114,3 +114,26 @@ The decision recorded here (decompose over monolith, compose via parent
 properties, no formal Visitor) was ratified by the Dart target — the same
 shape now lives in `src/Metano.Compiler.Dart/Bridge/` with no TypeScript
 contamination.
+
+## Thread-safety addendum (2026-05, ADR-0020)
+
+`TypeTransformer` now fans out the per-file-group loop across worker
+threads via `Parallel.For` (#21 / ADR-0020). The handler-decomposition
+shape is what makes that safe: every bridge consumes a
+`TypeScriptTransformContext` whose dictionaries / sets / maps are all
+read-only after the setup phase, and per-group mutable scratch state
+(`UsingAliases`, `NamedTypeRenames`) lives on `AsyncLocal<T>` slots
+that flow per worker. The only mutable shared sinks
+(`_diagnostics`, `TypeMappingContext.CrossPackageMisses`,
+`TypeMappingContext.UsedCrossPackages`) are now thread-safe at their
+declaration site (`lock` for the diagnostic list, `ConcurrentDictionary`
+for the cross-package collections).
+
+**Contract for new bridges**: a bridge introduced under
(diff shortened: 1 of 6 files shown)


danfma:
Addressed in bab8ba5.

@gemini-code-assist @copilot @chatgpt-codex-connector — three real correctness issues you all flagged:

1. `ConcurrentDictionary.Keys` is read-only — `CrossPackageMisses.Add(...)` would have thrown `NotSupportedException` the first time a cross-package miss fired. Fixed by introducing a small `ConcurrentHashSet<T>` wrapper around `ConcurrentDictionary<T, byte>` (the BCL ships no concurrent `HashSet`).
2. `TransformAll` constructed `new HashSet<string>()` / `new Dictionary<string, string>()` and passed them into the context, defeating the concurrent defaults. Fixed by dropping those allocations — the context now provisions concurrent collections itself.
3. The in-code comment in `TypeTransformer.TransformAll` referenced `Parallel.ForEach` while the code uses `Parallel.For`. Comment now matches the API.

ADR-0020 updated to describe the wrapper instead of the broken `Keys`-as-collection trick.

987 .NET tests pass, three back-to-back runs, no flakes; generated TS under `targets/` byte-identical.

PR #214, opened 2026-05-07
feat(compiler): incremental cache MVP — whole-build short-circuit

danfma:
## Summary

PR 3a/4 of the #21 + #18 thread. Adds a JSON cache file at `<outputDir>/.metano-cache.json` that pins per-source SHA-256, per-reference `length:lastWriteTimeUtc`, and per-output content hashes. `TranspilerHost.RunAsync` short-circuits the whole pipeline when every fingerprint matches and every cached output file still exists with matching content. Watch-mode (#18) noop ticks now cost only the load + extract phases.

- Cache opt-out via `--no-cache` (wired through `metano-typescript` and `metano-dart`)
- `--clean` and `--dry-run` skip both reading and writing
- Cache hits rehydrate `TranspileResult.Files` from disk so post-run consumers keep seeing every emitted artifact
- ADR-0021 documents the design + the deferred bits (per-type sig hashing, per-group skip → PR 3b)

### Roadmap (#21 + #18)

- [x] PR 3a — incremental cache: whole-build short-circuit (this PR)
- [ ] PR 3b — per-group skip via file-descriptor abstraction
- [ ] PR 4 — watch mode (#18)

### Known limitation (deferred to PR 3b)

Target-specific post-emit hooks (today only TypeScript's `PackageJsonWriter`) skip on cache hits because their inputs (`target.LastSourceFiles`, `target.LastEmitPackageName`, ...) live on the target instance and only get populated by `target.Transform`. The writer is idempotent (merge semantics) so an existing `package.json` survives correctly; the failure mode is "user deletes `package.json` between runs" — recovery is `--clean` or `--no-cache`. PR 3b folds the post-emit hook into the cache so this round-trips automatically.

## Test plan

- [x] 9 new unit tests for `TranspilationCache` round-trip, format-version mismatch, JSON corruption recovery, content hashing determinism, output-validity predicates, and rehydration
- [x] 996 total .NET tests pass (was 987)
- [x] Manual smoke: first run writes cache, second run hits, `touch` (no content change) stays a hit, real edit misses, deleted output file misses
- [x] Generated TS under `targets/` byte-identical
- [x] dual-agent review (compiler-man + bob) — addressed structural / readability findings; documented PackageJsonWriter limitation

Part of #21
Part of #18

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,10 @@ Thumbs.db
 *.tsbuildinfo
 bun.lock
 
+# Metano incremental cache — pinned next to generated outputs
+# (ADR-0021). Per-machine, never committed.
+.metano-cache.json
+
 # Local working plans (per CLAUDE.md workflow — private per-session notes,
 # not committed — the authoritative backlog lives in GitHub issues)
 tasks/
--- a/docs/adr/0021-incremental-cache.md
+++ b/docs/adr/0021-incremental-cache.md
@@ -1,0 +1,145 @@
+# ADR-0021 — Incremental cache: whole-build short-circuit
+
+**Status:** Accepted
+**Date:** 2026-05-07
+
+## Context
+
+Issue #21 asks for an incremental compilation cache so the transpiler
+skips work when nothing changed. Watch mode (#18) needs the same
+machinery: a file save that does not actually alter content (an IDE's
+save-on-focus-loss, a no-op formatter pass, a `touch`) should produce
+zero output instead of triggering a full re-emit.
+
+The dependency graph from PR 1 (#211 / ADR-0018) and the parallel
+TypeTransformer from PR 2 (#213 / ADR-0020) both expect a cache layer
+on top. The hard question is *granularity*: per-type (using sig
+hashes + the dependency graph) gives the finest skip but requires a
+file-descriptor abstraction so downstream stages (barrel generation,
+cyclic-import detection, file emit) can consume both freshly
+transformed and cached outputs uniformly.
+
+## Decision
+
+PR 3 ships the **whole-build short-circuit** as the MVP. Per-type
+skip ("PR 3b") is a follow-up that builds on top of this layer.
+
+The cache file lives at `<outputDir>/.metano-cache.json`. Shape:
+
+```json
+{
+  "formatVersion": 1,
+  "target": "TypeScript",
+  "sourceHashes":           { "<absolute .cs path>": "<sha256-hex>" },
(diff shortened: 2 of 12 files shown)


danfma:
Addressed in 71919b1.

@gemini-code-assist @chatgpt-codex-connector @copilot — five real correctness issues + two perf wins:

**Cache key was missing emit-affecting options** (gemini, codex, copilot — three independent flags): added `ITranspilerTarget.ConfigurationFingerprint` (TS target pins `NamespaceBarrels` + `StripInterfacePrefix`; Dart returns empty string), folded into the cache key alongside the host's `--file-prefix`. Schema bumped to `formatVersion: 2`. Smoke-tested: `--namespace-barrels` flip on a hit now misses.

**Path traversal in `ResolveOutputPath`** (copilot): `ValidateAndRehydrate` rejects rooted paths and `..` segments before opening the file. Two new tests (`ValidateAndRehydrate_RejectsRootedPath`, `_RejectsParentSegment`).

**`TryRead` accepted partially populated payloads** (copilot): explicit null-check on every required section after deserialization. New `TryRead_ReturnsNull_WhenRequiredSectionMissing` test.

**Prefix divergence on cache hit** (copilot): `RehydrateFiles` was returning prefixed on-disk content while `target.Transform` returns unprefixed. Replaced with `ValidateAndRehydrate(prefixBlock)` which strips the prefix when present. New `ValidateAndRehydrate_StripsPrefix_FromOnDiskContent` test.

**`--clean` writes the cache** (copilot): the read path skips on `--clean` (cache is wiped with the output dir), but the write path runs so the next run gets a fresh cache. Updated ADR-0021 to describe this behavior explicitly.

**Perf**:
- Source hashing now goes through Roslyn's `SourceText.GetChecksum()` instead of `tree.GetText().ToString()` — no full-file managed-string materialization (gemini).
- Output validation streams through `SHA256.HashData(Stream)` and rehydrates from a single disk pass per file (gemini, copilot — eliminates the double-read between `OutputsStillValid` + `RehydrateFiles`).
- `sourceHashes` and `referenceFingerprints` are computed lazily — `--no-cache` runs no longer pay the SHA bill (gemini, copilot).

**Deferred**: warning-count loss on cache hit (gemini high-priority). Documented as a known limitation in ADR-0021 already; the warnings from the previous successful run are not stored. Could be addressed in PR 3b if cheap.

Tests: 999 pass (was 996 — added null-section guard, prefix-aware rehydration, two path-traversal rejection cases).

PR #215, opened 2026-05-07
feat(compiler): --watch mode

### LINKED ISSUE #18: Watch mode (`--watch`) with FileSystemWatcher

danfma:
## Summary

PR 4/4 of the #21 + #18 thread. Closes #18.

Adds `--watch` to `metano-typescript` and `metano-dart`, backed by a target-agnostic `WatchHost` orchestrator (`src/Metano.Compiler/Watch/WatchHost.cs`):

- Single recursive `FileSystemWatcher` rooted at the project directory, filtering `.cs / .csproj / .props / .targets` and excluding `bin/` + `obj/` subtrees
- 250 ms quiet-period debounce + post-burst drain coalesces editor save bursts into a single recompile
- `Console.CancelKeyPress` → `CancellationTokenSource` for clean Ctrl+C exit
- Exceptions in the runOnce delegate log the full stack trace but do NOT stop the watcher — transient compile failures iterate, not abort
- The CLI builds a `RunOnce` delegate so target-specific post-emit (TypeScript's `PackageJsonWriter`) runs each tick

Combined with the incremental cache (ADR-0021), editor saves that do not actually change content cost only the load + extract phases. Real edits pay the parallel transform from ADR-0020.

### Roadmap (#21 + #18)

- [x] PR 4  — watch mode (this PR)
- [ ] PR 3b — per-group skip via file-descriptor abstraction (deferred follow-up)

## Test plan

- [x] 15 new test cases for `WatchHost.IsRelevant` (extension filter + `bin`/`obj` exclusion + mixed case + edge cases)
- [x] 1014 .NET tests pass (was 999)
- [x] Manual smoke on macOS: boot → initial transpile → real `.cs` edit triggers full recompile + package.json regenerated → `touch` triggers cache hit → Ctrl+C exits cleanly
- [x] Generated TS under `targets/` byte-identical
- [x] Dual review (compiler-man + bob) — addressed bin/obj exclusion, exception-stack logging, cancellation-flow simplification, debounce/drain extraction, naming improvements; deferred TS Commands `RunOnce` extract-method to a follow-up

Closes #18
Part of #21

### DIFF
--- a/docs/adr/0022-watch-mode.md
+++ b/docs/adr/0022-watch-mode.md
@@ -1,0 +1,111 @@
+# ADR-0022 — `--watch` mode
+
+**Status:** Accepted
+**Date:** 2026-05-07
+
+## Context
+
+Issue #18 asks for a long-running transpiler that re-emits whenever
+the source project changes, so iterative development does not pay the
+"open MSBuild project + create Roslyn compilation" cost on every save.
+The two prerequisites — fast incremental rebuilds (#21 / ADR-0021)
+and a parallel transformer (#21 / ADR-0020) — are now in place: the
+incremental cache absorbs no-op ticks (most editor saves) at the cost
+of just the load + extract phases, and a real change runs the
+transform across worker threads.
+
+## Decision
+
+Add `--watch` to both `metano-typescript` and `metano-dart`, backed
+by a target-agnostic `WatchHost` orchestrator under
+`src/Metano.Compiler/Watch/`.
+
+```text
+WatchHost.RunAsync(projectPath, runOnce, ct)
+```
+
+- `runOnce` is a delegate the CLI builds; it captures the same
+  pipeline the non-watch path runs (initial transpile + any
+  target-specific post-emit such as the TypeScript target's
+  `PackageJsonWriter`). This keeps `WatchHost` ignorant of target
+  shape and lets the watcher continue to drive every tick through
+  the existing CLI plumbing.
+- A single `FileSystemWatcher` is rooted at the project directory,
+  recursive, with `NotifyFilters.LastWrite | FileName | CreationTime
+  | Size`. The handler filters down to `.cs`, `.csproj`, `.props`,
+  and `.targets` extensions — everything else is editor noise (lock
+  files, swap files, `.tsbuildinfo`, `obj/`).
+- Events feed a `SemaphoreSlim` triggered after every relevant
+  change. The wait loop runs a 250 ms quiet-period debounce so an
+  IDE that fires a burst of events on save coalesces into a single
+  recompile, and then drains the semaphore so the next tick covers
+  the next burst.
+- `CancellationToken` exit: the CLI wires `Console.CancelKeyPress` to
+  a `CancellationTokenSource` so Ctrl+C triggers a clean shutdown.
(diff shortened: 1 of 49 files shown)


danfma:
Addressed in 2516c76 + 415760b.

@gemini-code-assist @chatgpt-codex-connector @copilot — eight findings across two passes:

**Correctness**:
- **Renamed handler now inspects both paths** (gemini high, codex P1, copilot): renaming a relevant input (`Foo.cs` → `Foo.txt`) or moving it under `bin/obj` triggers a recompile. Before, the rename was silently dropped.
- **`FileSystemWatcher.Error` handler** (gemini, copilot): buffer overflows surface as a yellow stderr warning telling the user to save again. The cache fingerprint catches up on the next save.
- **`Console.CancelKeyPress` is `-=` in `finally`** (gemini, both targets): the handler no longer leaks across CLI invocations and a Ctrl+C after watch exits cannot deref a disposed `CancellationTokenSource`.
- **Startup banner now lists every watched extension** (copilot): `.cs / .csproj / .props / .targets` instead of just `.cs / .csproj`.

**Thread-safety**:
- **`lastEventAtUtc` is now `long` ticks behind `Volatile.Read/Write`** (copilot): the watcher's background-thread writes never tear or get reordered against the consumer-loop reads.

**Behavior**:
- **Debounce sleeps for the *remaining* quiet period** (copilot): `WaitForQuietPeriod` now delays `DebounceMs - elapsed` instead of a fixed slice. Effective debounce is ~250 ms instead of drifting up to ~500 ms when an event lands right after a delay starts.

**Perf**:
- **`ContainsBuildArtifactSegment` walks the path via `MemoryExtensions.SplitAny` over a `ReadOnlySpan<char>`** (gemini): no per-event array allocation. `FileSystemWatcher` is chatty enough that `string.Split` showed up on the GC profile.

1014 .NET tests pass. Manual smoke on macOS: `--watch` boots with the updated banner; rename `Foo.cs` → `Foo.txt` triggers a recompile (errors surface for the missing type), rename back recovers via cache hit; `touch` (no content change) still hits the cache; Ctrl+C exits cleanly.

PR #216, opened 2026-05-11
feat(compiler): per-group skip foundation — type + closure signature hashing

danfma:
## Summary

Foundation for the per-group skip work (PR 3b). Adds two reusable, tested primitives in `src/Metano.Compiler/Caching/`:

- `IrTypeSignatureHasher.Hash(symbol)` — SHA-256 over a deterministic serialization of a Roslyn `INamedTypeSymbol` (identity, hierarchy, attributes, sorted member signatures, plus full declaring-syntax text so body edits invalidate even though the dep graph from ADR-0018 stays signature-only).
- `GroupClosureHasher.HashGroupClosure(groupFqns, graph, sigIndex)` — combines the per-type hashes for a group's transitive dependency closure into a single fingerprint. `BuildSignatureHashIndex` amortises the per-type hash across groups whose closures overlap.

The TypeTransformer integration (the user-visible per-group skip) is **explicitly deferred**. ADR-0023 explains why: downstream stages (`BarrelFileGenerator`, `CyclicReferenceDetector`) consume the full `TsSourceFile` AST, and a cache hit needs to hand them something that walks identically. Two designs are sketched in the ADR (stub `TsSourceFile` vs `IGeneratedFileSummary` interface); the integration PR picks one.

Why ship this alone:
- The hashers are pure functions — trivially testable, no shared state.
- The cache-invalidation contract is now frozen and reviewable. The integration PR becomes a wiring exercise, not a design exercise.
- The dep graph (PR 1) gets a concrete consumer; before this only ADR-0018 referenced it.

### Roadmap (#21 + #18)

- [x] PR 4 — watch mode (#215)
- [ ] **PR 3b — per-group skip (this PR is the foundation; integration is a follow-up)**

## Test plan

- [x] 9 new tests across the two hashers (1023 total .NET tests pass, was 1014)
- [x] `IrTypeSignatureHasher`: identical-source determinism, body edit flips, signature edit flips, attribute edit flips, base-type change flips
- [x] `GroupClosureHasher`: identical-closure determinism, seed-type edit flips, dependency edit flips (even when seed didn't change), edits to types **outside** the closure stay equal
- [x] Build green; `targets/` byte-identical
- [x] ADR-0023 + index updated

Part of #21

### DIFF
--- a/docs/adr/0023-per-group-skip-foundation.md
+++ b/docs/adr/0023-per-group-skip-foundation.md
@@ -1,0 +1,147 @@
+# ADR-0023 — Per-group skip foundation: type + closure signature hashing
+
+**Status:** Accepted (foundation only — integration deferred)
+**Date:** 2026-05-11
+
+## Context
+
+PR 3a (#214 / ADR-0021) shipped the whole-build short-circuit: if
+every source file, reference fingerprint, and output content matches
+the previous run, the entire pipeline is skipped. The follow-up
+asked for **per-group skip**: when only one file group's closure
+changes, only that group regenerates while the rest of the build
+reuses cached output.
+
+The dependency graph from PR 1 (#211 / ADR-0018) gives the
+"what types belong to a closure" answer. What's missing is the
+"did this closure change?" answer — a stable fingerprint that
+flips on body edits, signature edits, and attribute edits to
+anything inside the closure, and stays stable when an unrelated
+type elsewhere in the project moves.
+
+## Decision
+
+Ship the per-type and per-closure hashers as standalone, reusable
+primitives in `src/Metano.Compiler/Caching/`. The integration into
+`TypeTransformer.TransformGroup` is deferred to a follow-up PR
+once the file-descriptor abstraction is designed (see
+"Deferred" below).
+
+### `IrTypeSignatureHasher.Hash(INamedTypeSymbol)`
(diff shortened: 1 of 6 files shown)


danfma:
Addressed in 4cd1207.

@gemini-code-assist — three cache-key stability fixes:

- **Stable type display format**: every `ToDisplayString()` call in the hasher (base type, interfaces, member return / parameter types, attribute classes, `typeof(T)` arguments) now uses `SymbolDisplayFormat.FullyQualifiedFormat` instead of the default. Pins every namespace + containing-type prefix so short-name collisions across namespaces cannot tie unrelated types to the same hash, and the cache key cannot drift if Roslyn's default `ToDisplayString` changes.
- **Named attribute arguments**: `AppendAttributes` now folds in `NamedArguments` (sorted by key) alongside `ConstructorArguments`. Editing a named property like `[Import("Foo", "foo", Version = "2.0.0")]` invalidates the cache the same way a constructor-arg edit does.
- **`typeof(T)` constants**: `FormatConstant` handles `TypedConstantKind.Type` explicitly — `typeof` arguments now serialise via the stable format instead of the default `ITypeSymbol.ToString` (which can drift).

Added one test for the named-arg invalidation. 1024 .NET tests pass (was 1023).

PR #217, opened 2026-05-11
feat(compiler): per-group skip integration

### LINKED ISSUE #21: Incremental compilation and parallel TypeTransformer

danfma:
## Summary

Closes the per-group skip work from the #21 + #18 thread. PR 3a (#214) gave us the whole-build short-circuit; PR 3b (#216) gave us the per-type and closure hashers. This PR wires them into `TypeTransformer` so a single editor save regenerates only the affected groups while the rest reuse the disk content.

### Design (ADR-0024)

- **Stub `TsSourceFile` on cache hit**, **bypass the Printer** for cached paths, **skip `TransformGroup`** entirely.
- Cache file is target-specific (`<outputDir>/.metano-cache-groups-typescript.json`) and managed independently from PR 3a's target-agnostic cache. `--clean` wipes both together.
- `FileMetadataExtractor.BuildStub`: imports preserved fully, exports collapsed to `(name, typeOnly)` pairs and rebuilt as `TsConstObject` (value) or `TsInterface` (type-only) — the two `TsTopLevel` shapes `BarrelFileGenerator` matches on.
- `ITranspilerTarget.Transform` extended with optional `outputDir` + `filePrefix` parameters so the host plumbs both through without a side-channel property.

### Flow inside `TransformAll`

1. Compute per-group closure hashes (`BuildSignatureHashIndex` amortises per-type hashing across overlapping closures).
2. Load on-disk group cache.
3. Per group: **hit** → build stub via `FileMetadataExtractor`, capture the disk content (stripped of the host-applied file-prefix), return. **Miss** → run `TransformGroup`, extract metadata.
4. After loop: `Per-group cache: X/Y group(s) reused` summary if hits fired.
5. Write fresh cache.

### Roadmap (#21 + #18) — complete

- [x] PR 4 — watch mode (#215)
- [x] PR 3b — per-group skip foundation (#216)
- [x] PR 3c — per-group skip integration (this PR)

## Test plan

- [x] 1027 .NET tests pass (was 1024; +3 for `FileMetadataExtractor` round-trip)
- [x] Manual smoke on SampleIssueTracker (18 groups): edit one type → `Per-group cache: 17/18 group(s) reused from cache.`, byte-identical output for unchanged groups, regen for the touched group
- [x] All sample suites green; generated TS unchanged after fresh `--clean` rebuild
- [x] `.gitignore` updated for `.metano-cache-groups-*.json`

Closes #21

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,8 +23,9 @@ Thumbs.db
 bun.lock
 
 # Metano incremental cache — pinned next to generated outputs
-# (ADR-0021). Per-machine, never committed.
+# (ADR-0021, ADR-0024). Per-machine, never committed.
 .metano-cache.json
+.metano-cache-groups-*.json
 
 # Local working plans (per CLAUDE.md workflow — private per-session notes,
 # not committed — the authoritative backlog lives in GitHub issues)
--- a/docs/adr/0024-per-group-skip-integration.md
+++ b/docs/adr/0024-per-group-skip-integration.md
@@ -1,0 +1,170 @@
+# ADR-0024 — Per-group skip integration
+
+**Status:** Accepted
+**Date:** 2026-05-11
+
+## Context
+
+PR 3b (#216 / ADR-0023) shipped the hashers — `IrTypeSignatureHasher`
+for per-type fingerprints and `GroupClosureHasher` for per-group
+closure hashes. PR 3a (#214 / ADR-0021) shipped the whole-build
+short-circuit. Per-type sig hash plus the dep graph from PR 1
+(#211 / ADR-0018) gave us "did this group's closure change?", but
+the question we punted was **how to wire the skip into
+`TypeTransformer` without breaking downstream stages**
+(`BarrelFileGenerator`, `CyclicReferenceDetector`) that consume the
+full `TsSourceFile` AST.
+
+ADR-0023 sketched two designs (stub `TsSourceFile` vs
+`IGeneratedFileSummary` interface). This PR picks the **stub**
+design — surgical, minimal file churn, no abstraction refactor —
+and documents the constraints.
+
+## Decision
+
+### Cache file: `<outputDir>/.metano-cache-groups-typescript.json`
+
+Per-target group cache, keyed by `namespace/fileName`:
+
+```json
+{
+  "formatVersion": 1,
+  "groups": {
+    "Demo.Issues.Domain/issue": {
+      "closureHash": "<sha256>",
+      "files": [
+        {
+          "path": "issues/domain/issue.ts",
+          "imports": [
+            { "names": ["UserId"], "from": "#/shared-kernel/user-id" }
+          ],
+          "exports": [
+            { "name": "Issue", "typeOnly": false }
+          ]
+        }
+      ]
+    }
+  }
+}
+```
+
+The TypeScript target manages this file independently from PR 3a's
+target-agnostic cache; both files coexist next to the generated
+output so a single `--clean` wipes both.
+
+### Stub `TsSourceFile` on cache hit
+
+`FileMetadataExtractor`:
+- `Extract(TsSourceFile)` → builds `CachedFileMetadata` (path,
+  imports preserved with full `TsImport` shape, exports collapsed
+  to `(name, typeOnly)` pairs).
+- `BuildStub(CachedFileMetadata)` → reconstructs a stub
+  `TsSourceFile` with the original `TsImport` records plus minimal
+  export markers (`TsInterface` for type-only, `TsConstObject` for
+  value). Those are exactly the two `TsTopLevel` shapes the
+  downstream stages match on.
+
+The stub flows through `BarrelFileGenerator` and
+`CyclicReferenceDetector` identically to a fresh AST: barrel
+emission picks the right re-export form (value vs type-only),
+cycle detection walks the imports.
+
+### Bypass the Printer for cached files
+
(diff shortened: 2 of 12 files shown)


danfma:
Addressed in 31b76ec.

@chatgpt-codex-connector (P1) — `WriteGroupCache` on `--dry-run`: `TranspilerHost` now passes `null` for `outputDir` to `target.Transform` when `options.DryRun` or `options.NoCache` is set, so the TypeScript target's per-group cache no longer writes `.metano-cache-groups-typescript.json` on side-effect-free runs. Smoke-verified: `--dry-run` on SampleTodo emits no cache file.

@gemini-code-assist:
- **`_cachedFileContents` `ConcurrentDictionary`**: done, manual `lock` block dropped.
- **"Exported flag missing on stubs" (critical)**: false alarm — `TsInterface` and `TsConstObject` constructors default `Exported = true`, so `new TsInterface(name, [])` and `new TsConstObject(name, [])` produce exported stubs as required. Smoke confirmed barrels emit correctly.
- **"`TsModuleExport` not captured"**: by design. `BarrelFileGenerator.GetExportedName` doesn't read `TsModuleExport`, and `CyclicReferenceDetector` only walks `TsImport`. The actual on-disk content (which includes any `TsModuleExport` from the previous emit) is reused verbatim on cache hit — the stub only needs to satisfy the two downstream walkers. Smoke-verified with the `[ModuleEntryPoint]` sample (`SampleTodo.Service/Program.cs`): edit + rerun → `3/5 groups reused`, `program.ts` byte-identical (`dc7b93d5...`).
- **"`IsDefault` flag on `CachedExport`"**: moot for the same reason — default exports live in the cached on-disk content; the stub doesn't need to reconstruct them.

ADR-0024 documents the stub contract: any future downstream stage that starts reading other `TsTopLevel` shapes needs `FileMetadataExtractor` updated. That's the explicit extension point.

1027 .NET tests pass.

danfma:
Addressed in 8e0245f.

@copilot — 7 findings, all fixed:

- **outputDir/filePrefix gated when caching disabled**: `--no-cache` and `--dry-run` now pass `null` for outputDir to `target.Transform`; the TypeScript target loads no cache, writes none. `--clean` intentionally still writes a fresh cache (next run hits it) — same model as PR 3a.
- **Cache hits ignored target config / filePrefix**: `GroupCacheFile` schema bumped to formatVersion=2, now carries a `configurationFingerprint` field (target flags + host filePrefix). `LoadGroupCacheIfValid` discards the entire cache on mismatch. A `--strip-interface-prefix` toggle now invalidates everything.
- **Hit path trusted host validation that didn't happen**: on a normal transform run (PR 3a missed), the host did NOT verify disk content. Per-group hits now compute SHA-256 of the disk file and compare against the cached `ContentHash`. Smoke-verified: tampering `issue.ts` on disk causes a group miss and regenerates the file.
- **Path traversal**: `FileMetadataExtractor.IsSafeRelativePath` rejects rooted paths and `..` segments before opening cached paths (mirrors PR 3a's host-level check).
- **ComputePerGroupClosure ran unconditionally**: now skipped entirely when `CacheOutputDir` is null AND no cache was loaded — saves the per-type hash work on `--no-cache` / `--dry-run`.
- **Misleading XML doc**: `GroupCacheFile` doc rewritten to spell out the three conditions for re-use (config match, closure match, on-disk content hash match) instead of claiming host pre-verification.
- **WriteGroupCache wrote on error runs**: now skipped when the transform produced any `MetanoDiagnosticSeverity.Error` diagnostic — host aborts emit, so persisting entries pointing at unwritten files would seed stale hits.

Schema bump: formatVersion=2. Older v1 caches treated as miss and overwritten.

1027 .NET tests pass.

PR #233, opened 2026-05-15
feat(compiler): copy <MetanoAsset> static assets to generated output (#199)

### LINKED ISSUE #199: Copy assembly assets / resources to the generated package output

danfma:
## Summary
- Adds `<MetanoAsset Include="..." OutputPath="..." Language="..." />` MSBuild item that copies static asset files (schemas, fixtures, resources) into `MetanoOutputDir` alongside generated TS/Dart files.
- Repeatable `--asset SRC[=DEST]` CLI flag on both transpiler CLIs feeds `TranspilerHost.CopyAssetsAsync`, which runs after `EmitFilesAsync` and participates in the incremental cache via a new `AssetSourceHashes` section (cache format bumped to v3).
- Missing source, escape attempts (`..` segments or rooted destinations), and I/O failures raise MS0025 as **errors** so a broken manifest fails fast.
- `samples/SampleQueryableSqlite` now ships its DDL as a real `schema.sql` asset loaded from disk by `provider/db.ts`, replacing the inlined fixture string.

## Schema decisions
- **`OutputPath` (not `Output`)** — MSBuild reserves `Output` (MSB4066/MSB4118).
- **`Language` (not `Target`)** — same reason. `_MetanoTargetLanguage` defaults to `typescript`; downstream Dart integrations override before `MetanoTranspile` fires.
- **`<MetanoAsset>` defaults `<Language>all</Language>`** via `ItemDefinitionGroup`, so a bare `<MetanoAsset Include="x" />` applies to every target.

## Spec
- New FR-048 (`spec/04-functional-requirements.md`).
- Feature matrix row marked **Implemented**.
- New diagnostic MS0025 (`AssetCopyFailure`) catalogued.

## Test plan
- [x] `dotnet run --project tests/Metano.Tests/` — 1094 pass / 0 fail / 7 skipped (CS9282 pre-existing).
- [x] New unit tests:
  - `AssetFlagParserTests` — null / bare / `=` / trailing `=` / multiple `=` / blank entries.
  - `TranspilationCacheTests` — v3 round-trip, `ComputeAssetFingerprints` keying / skip-missing / destination-distinct / content-sensitive.
  - `TranspilerHostAssetCopyTests` — no-op empty list, mirrored happy path, explicit destination override, missing-source error, parent-escape error, rooted-destination error, multi-asset partial failure.
- [x] End-to-end: `dotnet build samples/SampleQueryableSqlite/SampleQueryableSqlite.csproj` copies `schema.sql` → `provider/schema.sql` next to the hand-written `db.ts`.
- [x] Incremental cache: rerun with no changes short-circuits via the new asset fingerprint; editing the asset content invalidates the cache.
- [x] CSharpier + Biome pre-commit / pre-push hooks all green (the 2 biome warnings on `sample-issue-tracker` are pre-existing).

## Follow-ups (not in this PR)
- Symlink-safe sandbox check (compiler-man flagged `Path.GetFullPath` does not resolve symlinks). Current check matches the destination string against the output prefix; defense-in-depth.
- Dart MSBuild target integration (when the Dart project gets its own `.targets`, it should override `_MetanoTargetLanguage=dart`).

Closes #199

### DIFF
--- a/samples/SampleQueryableSqlite/README.md
+++ b/samples/SampleQueryableSqlite/README.md
@@ -78,8 +78,9 @@ targets/js/sample-queryable-sqlite/
 ```
 
 The C# project sets `MetanoClean=false` so the hand-written TS providers under
-`src/provider/` survive between regenerations. Resource / asset copy from the
-C# project is tracked separately in #199.
+`src/provider/` survive between regenerations. The schema DDL ships as a
+`<MetanoAsset Include="schema.sql" OutputPath="provider/schema.sql" Language="typescript" />`,
+which the transpiler copies next to `provider/db.ts` on every build (FR-048).
 
 ## Running
 
@@ -112,6 +113,6 @@ bun test
 - #197 — `sample-queryable-arrays`, the in-memory provider this sample evolves.
 - #198 — generic visitor API + reusable walker (the natural extraction target
   once a few real providers exist).
-- #199 — copy assembly assets / resources to the generated package output (the
-  schema DDL would normally live as a `schema.sql` resource).
+- #199 — copy assembly assets / resources to the generated package output
+  (closed by this sample's `<MetanoAsset Include="schema.sql" />` wiring).
 - #200 — runtime `LINQ_STAGES` / `getStages` helper that powers introspection.
--- a/samples/SampleQueryableSqlite/SampleQueryableSqlite.csproj
+++ b/samples/SampleQueryableSqlite/SampleQueryableSqlite.csproj
@@ -17,5 +17,13 @@
     <ProjectReference Include="../../src/Metano/Metano.csproj" PrivateAssets="all" />
   </ItemGroup>
 
+  <ItemGroup>
+    <!-- Ship the schema next to the generated SQLite provider so db.ts
+         can load it at module init. OutputPath is relative to
+         $(MetanoOutputDir); Language="typescript" keeps the asset out of
+         a future Dart run. -->
+    <MetanoAsset Include="schema.sql" OutputPath="provider/schema.sql" Language="typescript" />
+  </ItemGroup>
+
   <Import Project="../../src/Metano.Build/build/Metano.Build.targets" />
 </Project>
--- a/samples/SampleQueryableSqlite/schema.sql
+++ b/samples/SampleQueryableSqlite/schema.sql
@@ -1,0 +1,13 @@
+-- Schema for the SampleQueryableSqlite product fixture.
+-- Shipped as a <MetanoAsset> so the same DDL is the single source of
+-- truth for both the C# side (if a query analyser ever wants to read
+-- it) and the generated TS bootstrap (provider/db.ts loads this file
+-- at module init).
+CREATE TABLE products (
+  id INTEGER PRIMARY KEY,
+  name TEXT NOT NULL,
+  display_name TEXT NOT NULL,
+  unit_price REAL NOT NULL,
+  stock_count INTEGER NOT NULL,
+  is_active INTEGER NOT NULL
+);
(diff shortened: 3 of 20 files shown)


danfma:
Addressed all review findings in bdc5901.

**P1 (chatgpt-codex-connector, Copilot) — missing-asset filter removed**
Dropped the `Exists('%(MetanoAsset.FullPath)')` filter in `Metano.Build.targets` so missing sources reach the host and surface as MS0025 errors instead of being silently dropped before the CLI invocation.

**P2 (chatgpt-codex-connector) — asset copy out of the zero-files guard**
`output.Files.Count == 0` no longer short-circuits when assets are declared. Asset-only runs now produce their files and diagnostics. `EmitFilesAsync` is skipped only when there are no generated files; `CopyAssetsAsync` always runs.

**Cache-hit asset validation (Copilot suppressed comment)**
Added `AreAssetCopiesIntact` to the cache-hit path: validates every declared asset still exists at its resolved destination with content matching the cached source fingerprint. Deleting an asset out-of-band now forces a full re-run that re-copies the file. Verified end-to-end: deleted `provider/schema.sql` → next build re-copies it.

**Stream-based hashing (Copilot, @gemini-code-assist)**
`ComputeAssetFingerprints` switched from `File.ReadAllBytes` to `SHA256.HashData(stream)`. No more full-file buffering.

**Cache key separator normalisation (Copilot, @gemini-code-assist)**
`ComputeAssetCacheKey` now normalises `\` → `/` in `RelativeDestination`, matching the host-side `NormalizeRelative`. `provider\schema.sql` and `provider/schema.sql` share a single cache row.

**MS0025 docstring (Copilot, @gemini-code-assist)**
"Surfaces the failure as a warning" → "as an error". `Output` → `OutputPath` to match the actual reserved-name-aware metadata.

**Windows / macOS case-insensitive paths (Copilot)**
`IsInsideOutputDir` now uses `OrdinalIgnoreCase` on Windows and macOS to match each platform's filesystem semantics; Linux keeps `Ordinal`.

**`Language=""` ambiguity (Copilot)**
Resolved by `ItemDefinitionGroup` defaulting `<Language>all</Language>` — an empty `Language` value is never observed because MSBuild applies the default. The condition's empty-string arm was redundant and got dropped earlier in the review pass.

All 1094 tests still pass. Sample regenerated cleanly.

